PR #2, opened 2019-03-29
Move hotmic to metrics.

tobz:
This is a WIP to move `hotmic` to its new home as `metrics`.  We're also taking some time to flesh out more of the `metrics-core` support.

Open issue(s):
- [x] high tail latencies for sends, presumably due to histogram changes


metrics/src/data/counter.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::data::ScopedKey;
+
+pub(crate) struct Counter {
+    data: HashMap<ScopedKey, u64, FnvBuildHasher>,
+}
+
+impl Counter {
+    pub fn new() -> Counter {
+        Counter {
+            data: HashMap::<ScopedKey, u64, FnvBuildHasher>::default(),
```
LucioFranco: you could prob drop this annotation :)

LucioFranco:
@tobz we prob also want to move that benchmark to be a real one based on like criterion?

PR #8, opened 2019-04-25
Add snapshot traits and expose exporters/recorders from metrics.

tobz:
We now expose all exporters and recorders via facade modules in the `metrics` crate, called `metrics::exporters` and `metrics::recorders`, respectively.  This means that the metrics crate itself has these as optional dependencies, which are included by the default set of features, and so can be turned off by consumers.

To curtail the issue of cyclical dependencies, we've also introduced three new traits: `Snapshot`, `SnapshotProvider`, and `AsyncSnapshotProvider`.

These traits let us represent `metrics::Controller` and `metrics::data::snapshot::Snapshot` in the exporter, allowing us to get around the cyclical dependency but also expose more flexibility and modularity.

### DIFF
--- a/metrics-core/src/lib.rs
+++ b/metrics-core/src/lib.rs
@@ -32,6 +32,7 @@
 //!
 //! Histograms are a convenient way to measure behavior not only at the median, but at the edges of
 //! normal operating behavior.
+use futures::future::Future;
 
 /// A value that records metrics.
 pub trait MetricsRecorder {
@@ -41,7 +42,7 @@ pub trait MetricsRecorder {
     /// as they are both a single value tied to a key.  From the perspective of a collector,
     /// counters and gauges usually have slightly different modes of operation.
     ///
-    /// For the sake of flexibility on the exportr side, both are provided.
+    /// For the sake of flexibility on the exporter side, both are provided.
     fn record_counter<K: AsRef<str>>(&mut self, key: K, value: u64);
 
     /// Records a gauge.
@@ -50,7 +51,7 @@ pub trait MetricsRecorder {
     /// as they are both a single value tied to a key.  From the perspective of a collector,
     /// counters and gauges usually have slightly different modes of operation.
     ///
-    /// For the sake of flexibility on the exportr side, both are provided.
+    /// For the sake of flexibility on the exporter side, both are provided.
     fn record_gauge<K: AsRef<str>>(&mut self, key: K, value: i64);
 
     /// Records a histogram.
@@ -58,4 +59,29 @@ pub trait MetricsRecorder {
     /// Recorders are expected to tally their own histogram views, so this will be called with all
     /// of the underlying observed values, and callers will need to process them accordingly.
     fn record_histogram<K: AsRef<str>>(&mut self, key: K, values: &[u64]);
+}
+
+/// A value that holds a point-in-time view of collected metrics.
+pub trait MetricsSnapshot {
+    /// Records the snapshot to the given recorder.
+    fn record<R: MetricsRecorder>(&self, recorder: &mut R);
+}
+
+/// A value that can provide on-demand snapshots.
+pub trait SnapshotProvider {
+    type Snapshot;
+    type SnapshotError;
+
+    /// Gets a snapshot.
+    fn get_snapshot(&self) -> Result<Self::Snapshot, Self::SnapshotError>;
+}
+
+/// A value that can provide on-demand snapshots asynchronously.
+pub trait AsyncSnapshotProvider {
+    type Snapshot;
+    type SnapshotError;
+    type SnapshotFuture: Future<Item = Self::Snapshot, Error = Self::SnapshotError>;
+
+    /// Gets a snapshot asynchronously.
+    fn get_snapshot_async(&self) -> Self::SnapshotFuture;
 }
--- a/metrics-core/src/lib.rs
+++ b/metrics-core/src/lib.rs
@@ -32,10 +32,11 @@
 //!
 //! Histograms are a convenient way to measure behavior not only at the median, but at the edges of
 //! normal operating behavior.
+use std::fmt::Display;
 use futures::future::Future;
 
 /// A value that records metrics.
-pub trait MetricsRecorder {
+pub trait Recorder {
     /// Records a counter.
     ///
     /// From the perspective of an recorder, a counter and gauge are essentially identical, insofar
@@ -62,15 +63,15 @@ pub trait MetricsRecorder {
 }
 
 /// A value that holds a point-in-time view of collected metrics.
-pub trait MetricsSnapshot {
+pub trait Snapshot {
     /// Records the snapshot to the given recorder.
-    fn record<R: MetricsRecorder>(&self, recorder: &mut R);
+    fn record<R: Recorder>(&self, recorder: &mut R);
 }
 
 /// A value that can provide on-demand snapshots.
 pub trait SnapshotProvider {
-    type Snapshot;
-    type SnapshotError;
+    type Snapshot: Snapshot;
+    type SnapshotError: Display;
 
     /// Gets a snapshot.
     fn get_snapshot(&self) -> Result<Self::Snapshot, Self::SnapshotError>;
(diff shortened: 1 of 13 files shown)


metrics-core/src/lib.rs
```diff
@@ -59,3 +60,28 @@ pub trait MetricsRecorder {
     /// of the underlying observed values, and callers will need to process them accordingly.
     fn record_histogram<K: AsRef<str>>(&mut self, key: K, values: &[u64]);
 }
+
+/// A value that holds a point-in-time view of collected metrics.
+pub trait MetricsSnapshot {
```
LucioFranco: prob can drop the word metric from here?

metrics-core/src/lib.rs
```diff
@@ -59,3 +60,28 @@ pub trait MetricsRecorder {
...
+    /// Records the snapshot to the given recorder.
+    fn record<R: MetricsRecorder>(&self, recorder: &mut R);
+}
+
+/// A value that can provide on-demand snapshots.
+pub trait SnapshotProvider {
+    type Snapshot;
```
LucioFranco: shouldn't this be `type Snapshot: MetricSnapshot`?

metrics-exporter-log/src/lib.rs
```diff
@@ -66,23 +78,16 @@ where
...
-        },
-        Err(e) => log!(Level::Error, "failed to capture snapshot: {}", e),
+                controller.get_snapshot_async()
+                    .and_then(move |snapshot| {
+                        snapshot.record(&mut recorder);
+                        let output = recorder.into();
+                        log!(level, "{}", output);
+                        Ok(())
+                    })
+                    .map_err(|e| log!(Level::Error, "failed to get snapshot: {}", e))
```
LucioFranco: `error!`?

metrics/Cargo.toml
```diff
@@ -21,14 +21,23 @@ debug = true
 opt-level = 3
 lto = true
 
+[features]
+default = ["exporters", "recorders"]
+exporters = ["metrics-exporter-log"]
```
LucioFranco: again here prob can drop all the word metrics from these features?

tobz: For the features, they need to match the name of the package.  This utilizes Cargo's support for [defining a feature name to include an optional dependency](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section).

metrics-exporter-log/src/lib.rs
tobz: Yeah, that's fair.  I was just following the `log!` usage above it.  I'll change it for succinctness, though.

metrics/Cargo.toml
LucioFranco: oh im blind :P

metrics-core/src/lib.rs
tobz: Good call! This lets me drop all those constraints on the exporter. :D

metrics-core/src/lib.rs
tobz: Yeah, it's probably reasonable to drop `Metrics` from both the snapshot trait _and_ recorder trait.  I'll do that.

tobz:
@LucioFranco This is ready for another pass.

metrics-core/src/lib.rs
```diff
@@ -59,3 +61,28 @@ pub trait MetricsRecorder {
...
+/// A value that holds a point-in-time view of collected metrics.
+pub trait Snapshot {
+    /// Records the snapshot to the given recorder.
+    fn record<R: Recorder>(&self, recorder: &mut R);
+}
+
+/// A value that can provide on-demand snapshots.
+pub trait SnapshotProvider {
+    type Snapshot: Snapshot;
+    type SnapshotError: Display;
```
LucioFranco: `: Error` bound instead? or do you even need to bound this?

metrics-exporter-log/src/lib.rs
```diff
@@ -7,35 +7,34 @@
...
+    controller: C,
     recorder: R,
     level: Level,
 }
 
-impl<R> LogExporter<R>
+impl<C, R> LogExporter<C, R>
 where
-    R: MetricsRecorder + Clone + Into<String>
+    C: SnapshotProvider + AsyncSnapshotProvider,
```
LucioFranco: can we move these bounds to `turn` and `into_future`?

metrics/src/control.rs
```diff
@@ -46,16 +53,39 @@ impl Controller {
...
-            .map_err(|_| SnapshotError::ReceiverShutdown)
-            .map(move |_| rx)
+            .map(move |_| SnapshotFuture::Waiting(rx))
+            .unwrap_or(SnapshotFuture::Errored(SnapshotError::ReceiverShutdown))
+    }
+}
+
+pub enum SnapshotFuture {
+    Waiting(oneshot::Receiver<Snapshot>),
+    Errored(SnapshotError),
```
LucioFranco: not totally sure you need this variant? Can't you just poll the future and check its internal error?

metrics-core/src/lib.rs
tobz: So, for the non-async stuff, it has to have a bound so we can display it.  `Error` could probably work fine here, since it requires `Display` and `Debug` to begin with.

But yeah, if we don't bound it, then we can't log the error.

tobz: Sorry, that's to say.... we don't need to bound it necessarily..  I suppose we could defer the bound to the users -- i.e. the exporters -- who make the decision on whether or not they're going to log/emit/etc any errors they get.

I'll move the bound there, and change `SnapshotError` (the actual struct type, not the associated type in the trait) to impl `Error` and then bound on `Error` for the log exporter, just to show the proper way to do it for future exporter impls.

metrics-exporter-log/src/lib.rs
tobz: Is there any practical reason to do this?  It seems like it's 50/50 on whether types constraint at the type level or at the function level; I notice Tokio/Tower prefer the some of "late binding" constraint approach.

metrics-core/src/lib.rs
LucioFranco: Yeah, this bound should probably be moved to who ever takes this impl.

metrics-exporter-log/src/lib.rs
LucioFranco: well, in this case, someone could pass a non-async one and still use this.

metrics/src/control.rs
tobz: It handles the case when there's an error even sending the command to the receiver i.e. the receiver thread is dead so nobody can receive our request to generate a snapshot.

Thinking about it, though... maybe if the sender of the oneshot pair we generate drops because it was never able to be sent, then the receiver of the pair would implicitly error as the sender dropped without sending a value... so we could likely drop this variant and switch to a pure struct.  I'll play around with that.

metrics-exporter-log/src/lib.rs
tobz: Good point, yeah, letting people write a non-async provider if they don't need async and vise versa.  Will do this.

metrics-core/src/lib.rs
tobz: Yep, makes sense.  Will do.

metrics/src/control.rs
tobz: Thinking through this a little more: I think I want to leave this as is.

Essentially, two errors can happen:
- the receiver is shut down (thread is stopped, thread panics, whatever)
- an error occurs specifically during generating (and sending back) the snapshot

If we rely on the `oneshot::Receiver` to give us back an error, the only error, we wouldn't know if the receiver was shut down or if an error occurred during the generation of the snapshot.  I'd like to keep it as is because it lets us inform the user if the receiver is shutdown, which, more likely than not, will primarily be when they forget to start it.

It's just a nice little differentiation, IMO, to give a more useful error message between the two cases without any performance loss or loss of ergonomics.

PR #13, opened 2019-05-28
Entirely remove the event loop and switch to pure atomics.

tobz:
Originally, metrics (and `hotmic` before it was converted) was based on an event loop that centered around `mio`'s `Poll` interface with a custom channel to read and write metrics to.  That model required a dedicated thread to run to poll for writes, and ingest them, managing the internal data structures in turn.

Eventually, I rewrote that portion to be based on `crossbeam-channel` but we still depended on a background thread to pop samples off the channel and process them.

Instead, we've rewritten the core of metrics to be based purely on atomics, with the caveat that we still do have a background thread.

Instead of a single channel that metrics are funneled into, each underlying metric becomes a single-track codepath: each metric is backed by an atomic structure which means we can pass handles to that storage as far as the callers themselves, eliminating the need to funnel metrics into the "core" where they all contend for processing.

Counters and gauges are now, effectively, wrapped atomic integers, which means we can process over 100 million counter/gauge updates per core. Histograms are based on a brand-new atomic "bucket" that allows for fast, unbounded writes and the ability to snapshot at any time.

The end result is that we can process a mixed workload (counter, gauge, and histogram) at sample rates of up to 30 million samples per second per core, with p999 ingest latencies in the low hundreds of nanoseconds.  Taking snapshots also now avoids stalling the event loop and driving up tail latencies for ingest, and writers can proceed with no interruption.

There is still a background thread that is part of quanta's new "recent" time support, which allows a background thread to incrementally update a shared global time, which can be accessed more quickly than grabbing the time directly.  For our purposes, only histograms need the time to perform the window upkeep inherent to the sliding windows we use, and the time they need can be far less precise than what quanta is capable of.  This background thread is spawned automatically when creating a receiver, and drops when the receiver goes away.  By default, it updates 20 times a second performing an operation which itself takes less than 100ns, so all in all, this background thread should be imperceptible, performance-wise, on all systems*.

On top of all of this, we've embraced the natural pattern of defining metrics individually at the variable/field level, and added supported for proxy types, which can be acquired from a sink and embedded as fields within your existing types, which lets you directly update metrics with the ease of accessing a field in an object.  Sinks still have the ability to have metrics pushed directly into them, but this just opens up more possibilities.

(*) famous last words


metrics/src/builder.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// buckets, etc.
+    ///
+    /// Acquiring the current time is fast compared to most operations, but is a significant
+    /// portion of the other time it takes to write to a histogram, which limits overall throughput
+    /// under high load.
+    ///
+    /// We spin up a background thread, or the "upkeep thread", which updates a global time source
+    /// that the read and write operations exclusively rely on.  While this source is not as
+    /// up-to-date as the real clock, it is much faster to access.
+    UpkeepFailure,
```
LucioFranco: we should add a 

```rust
#[doc(hidden)]
_NonExahustive
```

metrics/src/builder.rs
```diff
@@ -0,0 +1,77 @@
...
+        match self {
+            BuilderError::UpkeepFailure => write!(f, "failed to spawn quanta upkeep thread"),
+        }
+    }
+}
+
+/// Builder for [`Receiver`].
+#[derive(Clone)]
+pub struct Builder {
+    pub(crate) histogram_window: Duration,
```
LucioFranco: do you actually need this to be pub at all?

metrics/src/common.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+impl MetricValue {
+    fn new(state: ValueState) -> Self {
+        MetricValue {
+            state: Arc::new(state),
+        }
+    }
+
+    pub(crate) fn new_counter() -> Self {
```
LucioFranco: miiiiiiiiiight make sense to put all these pub(crate) impls into their own `impl` block?

metrics/src/common.rs
```diff
@@ -0,0 +1,235 @@
...
+                let stream = inner.snapshot();
+                ValueSnapshot::Histogram(stream)
+            }
+        }
+    }
+}
+
+/// Trait for types that represent time and can be subtracted from each other to generate a delta.
+pub trait Delta {
+    fn delta(&self, other: Self) -> u64;
```
LucioFranco: we should add `warn` docs cuz this needs a doc comment 😄

metrics/src/control.rs
```diff
@@ -1,53 +1,49 @@
...
 use metrics_core::{AsyncSnapshotProvider, SnapshotProvider};
 use std::error::Error;
 use std::fmt;
-use tokio_sync::oneshot;
+use std::sync::Arc;
 
-/// Error conditions when retrieving a snapshot.
+/// Error during snapshot retrieval.
 #[derive(Debug, Clone)]
 pub enum SnapshotError {
```
LucioFranco: same non exahustive stuff as above to support forward compat

metrics/src/data/histogram.rs
```diff
@@ -1,212 +1,229 @@
...
+                // while we're doing the clear.
+                self.buckets[new_index].clear();
+
+                // Since another write could outrun us, just do a single CAS.  99.99999999% of the
+                // time, the CAS will go through, because it takes nanoseconds and our granularity
+                // will be in the hundreds of milliseconds, if not seconds.
+                self.next_upkeep.compare_and_swap(
+                    next_upkeep,
+                    next_upkeep + self.granularity,
+                    Ordering::AcqRel,
```
LucioFranco: why the AcqRel not SeqCst?

metrics/src/receiver.rs
```diff
@@ -1,220 +1,73 @@
...
-    counter: Counter,
-    gauge: Gauge,
-    thistogram: Histogram,
-    vhistogram: Histogram,
-
+    metric_registry: Arc<MetricRegistry>,
+    scope_registry: Arc<ScopeRegistry>,
     clock: Clock,
-    scopes: Arc<Scopes>,
+    _upkeep_handle: UpkeepHandle,
```
LucioFranco: is this still needed?

metrics/src/builder.rs
tobz: It's pub so that `Receiver` can access it when we call `Receiver::from_builder(builder_obj)`, but I'm open to suggestions.  I did start the process of creating `MetricsConfiguration` where those values are stored so that we can pass around a single struct instead of potentially multiple values needed when creating a metric.  Just not sure if `Builder` should generate that and pass it into `Receiver::from(..)`.

I'd be fine with making that change, although they'd still have to be either public fields or getters.

Thoughts?

metrics/src/builder.rs
tobz: What does this buy us?

metrics/src/receiver.rs
tobz: The `UpkeepHandle` is essentially a newtype around `std::thread::JoinHandle`, and so we need to hold it to keep the background Quanta time thread running.  I only prefixed it to get the compiler to shut up about it.

metrics/src/builder.rs
LucioFranco: meh im fine with it that way

metrics/src/builder.rs
LucioFranco: if someone decides to match the error enum and we add new variants it wont break their code. Pretty much it forces them to have a catch all arm.

metrics/src/receiver.rs
LucioFranco: 👍

metrics/src/builder.rs
tobz: I ended up creating a `Configuration` type that will hold all of this, and have the pub fields.  The `Builder` will just be the fluent API while `Configuration` can be cloned/copied/passed around for anything that needs the config data.

metrics/src/builder.rs
tobz: Done!

metrics/src/common.rs
tobz: I made all the inner methods regular `pub` since the type itself is `pub(crate)`, and I also dropped the `new_` prefix on some of the methods to make it a little cleaner.

metrics/src/common.rs
tobz: I added `#![deny(missing_docs)]` to really make sure we don't slip up... and I added a doc comment for `delta`. :)

metrics/src/control.rs
tobz: Done.

metrics/src/config.rs
```diff
@@ -2,16 +2,19 @@ use crate::Builder;
 use std::time::Duration;
 
 /// Holds the configuration for complex metric types.
-pub(crate) struct MetricConfiguration {
+#[derive(Clone)]
```
LucioFranco: debug?

tobz: Done.

PR #17, opened 2019-05-29
Update our README.

tobz:
It's time to get rid fo the caveat emptor and fully embrace our new crate home.

This README likely also needs to be copypasta'd into the actual `metrics` crate in the workspace so they're basically both the same.


tobz:
We also need to make sure we update the performance section either when the atomics PR lands, or if it lands before this PR does, then we should do it here.

Issue #18, opened 2019-05-29
log-style API proposal


kazimuth:
I'm excited people are working on a metrics library for rust! I've been wanting one for a while. Your reddit post said you were okay with big API design ideas, so here are some i've been noodling on :)

The design I was contemplating is something like [`log`](https://github.com/rust-lang-nursery/log), where there's a tiny facade crate with simple macros that library authors can use freely, that calls into a global trait object for functionality; binaries can then install an implementation that does what they want.

E.g., library code can do something like:

```rust
use metrics::gauge;

gauge!(request_time, 123);
```

(Or, if we want to get fancy, you could also more metadata, like prometheus-style tags and opencensus-style units:)
```rust
gauge!(request_time, 123ns, endpoint="/app/thing", method=http::POST);
```

You could also have other macros to record counters, histograms, time blocks of code, etc. But it's all just simple macros.

Then, users can install a metrics implementation:

```rust
fn main() {
   // note: NOT the core metrics crate!
   magic_metrics_library::setup()
     .with_recorder(PrometheusRecorder::new())
     .with_exporter(HttpExporter::new())
     .filter("hyper") // don't include hyper metrics...
     .install();

   start_http_server();
}
```

And boom, they're done! That's the whole user facing API. Users don't need to know anything more about the internal details of the library. This is especially nice because it means library authors can just add metrics, fire-and-forget style; the core crate is so small that it adds very little compile-time cost, and also practically no runtime cost. (`log` without a logger installed has a 1ns overhead per-call, by my measurements.) As a result, when a user installs a metric exporter, they'll hopefully have a load of metrics available for free; assuming the facade crate gains adoption.

Now, the implementation. Basically, the core metrics crate contains a core `Metrics` trait that allows metrics to be recorded -- the API surface looks similar to the `Recorder` trait, but its responsibilities are broader; it's responsible for the whole metric recording process. It also contains a global, atomically initialized `static METRICS: Box<dyn Metrics>` trait object. Then, each `gauge!`, `counter!`, or `histogram!` call simply calls a method on this trait object. (You can look at `log` to see how this works.)

Note that we don't even have SnapshotProvider / Exporter / etc here! Those are super useful, but they can be moved to a crate for backend implementors -- we don't want their compile time cost in the facade crate. (Note: I've been calling the facade crate `metrics` in the style of `log`, but it could be `metrics_core` instead.)

[...]

PR #20, opened 2019-05-30
Add metrics-facade crate to provide global macros.

tobz:
Based on #18, this is an experimental implementation of providing global macros -- `counter!`, `gauge!`, `timing!`, `value!` -- in the `log` style.

These are provided through a new crate -- `metrics-facade` -- that, like `log`, defines its own trait for an object that can record metrics.  There's a global recorder instance that can be overridden at runtime, and is used as the recorded/sink for all macro-based calls.

We also added support in `Receiver` with a new method -- `pub fn install(self)` -- that allows you to take a `Receiver` you built and then install it into the global recorder slot, again, similar to how crates like `env_logger` work with the `log` facade.

The performance is currently pretty meh: even based on the atomics branch, updating a metric is roughly 3x slower, and we're using thread locals to store a reference to a dedicated `Sink` which should have a locally cached handle for the metrics.  There's probablyyyy something we do along the lines of the callsite-based "token" registration, not too dissimilar from scope IDs to get from a slow hashmap lookup to an index-into-a-vector lookup.

/cc @kazimuth

### DIFF
--- a/metrics-facade/src/lib.rs
+++ b/metrics-facade/src/lib.rs
@@ -1,12 +1,160 @@
+//! A lightweight metrics facade.
+//!
+//! The `metrics-facade` crate provides a single metrics API that abstracts over the actual metrics
+//! implementation.  Libraries can use the metrics API provided by this crate, and the consumer of
+//! those libraries can choose the metrics implementation that is most suitable for its use case.
+//!
+//! If no metrics implementation is selected, the facade falls back to a "noop" implementation that
+//! ignores all metrics.  The overhead in this case is very small - an atomic load and comparison.
+//!
+//! # Use
+//! The basic use of the facade crate is through the four metrics macros: [`counter!`], [`gauge!`],
+//! [`timing!`], and [`value!`].  These macros correspond to updating a counter, updating a gauge,
+//! updating a histogram based on a start/end, and updating a histogram with a single value.
+//!
+//! Both [`timing!`] and [`value!`] are effectively identical in so far as that they both translate
+//! to recording a single value to an underlying histogram, but [`timing!`] is provided for
+//! contextual consistency: if you're recording a measurement of the time passed during an
+//! operation, the end result is a single value, but it's more of a "timing" value than just a
+//! "value".  The [`timing!`] macro also has a branch to accept the start and end values which
+//! allows for a potentially clearer invocation.
+//!
+//! ## In libraries
+//! Libraries should link only to the `metrics-facade` crate, and use the provided macros to record
+//! whatever metrics will be useful to downstream consumers.
+//!
+//! ### Examples
+//!
+//! ```rust
+//! #[macro_use]
+//! extern crate metrics_facade;
+//!
+//! # use std::time::Instant;
+//! # pub fn run_query(_: &str) -> u64 { 42 }
+//! pub fn process(query: &str) -> u64 {
+//!     let start = Instant::now();
+//!     let row_count = run_query(query);
+//!     let end = Instant::now();
+//!
+//!     timing!("process.query_time", start, end);
+//!     counter!("process.query_row_count", row_count);
+//!
+//!     row_count
+//! }
+//! # fn main() {}
+//! ```
+//!
+//! ## In executables
+//!
+//! Executables should choose a metrics implementation and initialize it early in the runtime of
+//! the program.  Metrics implementations will typically include a function to do this.  Any
+//! metrics recordered before the implementation is initialized will be ignored.
+//!
+//! The executable itself may use the `metrics-facade` crate to record metrics well.
+//!
+//! ### Warning
+//!
+//! The metrics system may only be initialized once.
+//!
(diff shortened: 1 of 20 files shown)


kazimuth:
huh, how much slower is 3x, 600ns per call? i wonder where the cost is coming from. i wouldn't think accessing a thread-local was that expensive, even going through a RefCell.

kazimuth:
Update: It turns out most of the extra cost can be fixed with API changes!

- There's a lot of overhead from allocating Strings for each call. This can be fixed by having `MetricReceiver` take &'static str, so there's no need to box. (We could maybe offer a separate API for dynamic metric names? Is that even a use case for the macro API?)
- There's also some overhead from redundant checks on looking up the thread-local Sink, those're fixable through a little bit of hacking.
- ~70% of the remaining runtime cost comes from the hashmap lookup of the metric. That should be fixable using the token-based API I sketched.

In the end I believe the macro-based api will still add 20-30% overhead over the non-macro API. The cost of the dynamic call + thread-local access isn't really avoidable. We're still in the 100ns range timing wise though so I don't think that's a dealbreaker.

I'd be happy to implement these changes in a branch if you want.

tobz:
Not sure why responding via e-mail didn't show up here, but:

Yeah, if you want to fork my branch and play with it, that's fine.  I use the `benchmark` example as my gauge, specifically the "sender latency" histogram at the end of the run.

I'm not 100% sold on always requiring a static string, but it's definitely a reasonable thing to toy with right now. Like fundamentally, yes, it will be a hard-coded string 99% of the time.

The other thing is that I think it probably only makes sense if we append the metric name to the name of the current module/function, since we lose scoping here and having callsites where every metric "name" looks like `myapp_module_thing_measurement` instead of just `measurement` feels bad.  It would also mimic the `log` facade and allow for ubiquitous addition to any create, like you mentioned.

tobz:
Yeah, if you want to fork my branch and play with it, that's fine.  I use
the `benchmark` example as my gauge, specifically the "sender latency"
histogram at the end of the run.

I'm not 100% sold on always requiring a static string, but it's definitely
a reasonable thing to toy with right now. Like fundamentally, yes, it will
be a hard-coded string 99% of the time.

The other thing is that I think it probably only makes sense if we append
the metric name to the name of the current module/function, since we lose
scoping here and having callsites where every metric "name" looks like
`myapp_module_thing_measurement` instead of just `measurement` feels bad.
It would also mimic the `log` facade and allow for ubiquitous addition to
any create, like you mentioned.

tobz:
I didn't profile it yet but I'm measuring the cost of a "round" which is
sending three metrics and measuring the time.  Each call has the hit the
TLS, extract the sink, and traverse the sink's local metric cache, which is
a hashmap.

It also has to copy the metric name string key; apparently you can't have
generics in a trait if you want to do `&'static Trait`, and normally the
sink takes `Into<Cow<'static, str>>` to make static string metric names as
fast as possible.

The second one might be fixable, I was just trying to make the compiler
happy about the lifetimes.  First one I think just needs some cache love
related to your `CallsiteMetadata` idea, I think.

metrics-facade/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub fn set_boxed_recorder(recorder: Box<MetricsRecorder>) -> Result<(), SetRecorderError> {
+    set_recorder_inner(|| unsafe { &*Box::into_raw(recorder) })
+}
+
+fn set_recorder_inner<F>(make_recorder: F) -> Result<(), SetRecorderError>
+where
+    F: FnOnce() -> &'static MetricsRecorder,
+{
+    unsafe {
```
LucioFranco: what is actually unsafe here...?

tobz: Mutable static.

    error[E0133]: use of mutable static is unsafe and requires unsafe function or block
      --> metrics-facade/src/lib.rs:43:17
       |
    43 |                 RECORDER = make_recorder();
       |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^ use of mutable static
       |
       = note: mutable statics can be mutated by multiple threads: aliasing violations or data races will cause undefined behavior

LucioFranco: ah....

metrics-facade/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// Records a histogram.
+    ///
+    /// Recorders are expected to tally their own histogram views, so this will be called with all
+    /// of the underlying observed values, and callers will need to process them accordingly.
+    ///
+    /// There is no guarantee that this method will not be called multiple times for the same key.
+    fn record_histogram(&self, key: Key, value: u64);
+}
+
+struct NopRecorder;
```
LucioFranco: isnt it technically `NoopRecorder`? 😄

tobz: Good call.  Fixed.

Issue #24, opened 2019-06-13
Consider renaming metrics-core + metrics-facade


kazimuth:
This is just a suggestion - no problem if there isn't interest! I understand names can be a very touchy issue for bikeshedding :)

In my opinion the goal of `metrics` should be to achieve similar ecosystem penetration to `log` -- it should be a no-brainer for all rust crates to add metrics, since those metrics will add very little compile-time and run-time cost. I'd love to be able to turn on a prometheus exporter and already have most of my metrics predefined, or add a flag to my CLI app and instantly get a dump of execution telemetry. For that to happen, the metrics crate should provide a dead-simple API and an excellent documentation experience. And the most important part of documentation is names.

Currently, however, the names of the involved crates are a little confusing. The `metrics` crate is complex and opinionated, whereas basic functionality required for ecosystem interop is relegated to the crates `metrics-core` and `metrics-facade`. If the crates were released as-is, it'd be like if `fern` was named `log` and `log` was named `log-core`. Not the end of the world, but a little confusing, especially for what's aiming to become a fundamental piece of infrastructure.

So I'd like to suggest combining `metrics-core` + `metrics-facade` and renaming them to `metrics`, and renaming the current `metrics` crate to something like `metrics-runtime`, or even back to `hotmic`. This would also avoid privileging `metrics` over e.g. `metered` and `dipstick` -- since they'd all comply with the same basic interop stuff, they'd be on a level playing field.

I do think `metrics-facade` needs more work before it's ready for prime-time; I've got a PR partway done that offers some decent perf improvements, and I think it should be fairly straightforward to get `no_std` support fully functional as well.

Thoughts?

LucioFranco:
As a small counter point, tokio operates similar to how metrics is currently setup. The `tokio` crate is batteries included which I think was kinda the goal with metrics, its just a crate the reexports the smaller individual crates. Really the important stuff is the glue that lives in `metrics-core` but those could be moved to their own crates that are small.

tobz:
Yeah,

I can see where you're coming from, but I'm also of two minds here.

The `metrics` crate is high visibility, which is precisely what we wanted when we asked to take it over from the previous maintainer.  Further, it's always been a crate with an actual implementation.

That being said, from a "nice to have" perspective, it would be nice to `metrics` match `log` and be the facade crate.

[...]

Issue #23, opened 2019-06-13
Refactor `Key` to be a struct


kazimuth:
Refactoring idea: redefine Key to be similar to `common::MetricIdentifier`, and just use that everywhere from `metrics-core` up:

```rs
#[derive(Clone, Copy, PartialEq, Eq, Hash)]
struct Key<'_> {
   kind: MetricType,
   name: Cow<str>,
   scope: Cow<str>,
   labels: Cow<[(Cow<str>, Cow<str>)]>,
}

#[derive(Clone, Copy, PartialEq, Eq, Hash)]
enum MetricType {
    Counter,
    Gauge,
    Histogram
}
```

This would make label support easy + turn a bunch of APIs from `record_xyz(name, value)` to just `record(Key {name, value, scope, labels})`. It would also hopefully make defining the `Ingest` trait really easy (#21).

tobz:
Yeah, having `Key` as a struct would make easier to collapse the trait that facade recorders need to implement with the idea of a generic ingest trait, since you can't have generics in static mutables, and that's the main reason we have `metrics-facade` defining its own trait.

Would need to sketch this out in a PR, I think, to get a better feel for it.

tobz:
This happened in #27.

Issue #11, opened 2019-04-30
Is the recorder cloning approach really the best we can do?


tobz:
Context:

We made the decision to split exporters and recorders into distinct types.  This means that instead of a single crate/type that does, say, Prometheus over HTTP, there would be an HTTP _exporter_ and a Prometheus _recorder_, separating the transport and presentation concerns.  This is all fine and good.

Stemming from this fact, some exporters will naturally want the recorder they're using to give them the output of a snapshot in a particular format, such as a `String`.  This is also fine.

The problem comes in where our `Snapshot` trait exists: a snapshot should be able to take a `Recorder` and write all of its data into it.  For our existing recorders, they make themselves `Clone`, with a specialized implementation to keep around things like histogram configuration, but zero out other internal storage.  This is because we wouldn't want to record two or more snapshots into a single recorder, as we would totally mess up the values by duplicating or overwriting.  On top of that, our current exporter -- the log exporter -- also expects that the recorder will be `Into<String>` to get the string output... so we always have to clone the original recorder that was passed in since we consume it at the end of recording the snapshot.

There's a few potential alternatives:
- a type/trait for a recorder "builder", where recorders configure a builder, hand _that_ off to the exporters, and the exporters build a new recorder instance for every snapshot
- a new trait that works similarly to `Clone` but is named to indicate that you're getting a "fresh" clone, so that people don't think they're getting a _true_ cloned object
- a new trait that recorders would implement that mutably gets the "value" of the recorder (so, the string output of everything recorded into it at that point in time) and gives it a chance to clear its internal state... something like `Flush` or `Finalize`

Right now, `Clone` is _fine_, but I think we're sort of abusing it and could potentially do better.

tobz:
After chatting about it in #27, I'm going to take the builder pattern approach to solve this issue.

tobz:
After discussing it in #27, I'm going to take the builder pattern approach here.

PR #27, opened 2019-06-23
Add label support + rename crates.

### LINKED ISSUE #5: Investigate attaching tags to metrics.
Naive thought is that we switch to a composite metric key internally -- an enum, really -- where we have untagged and tagged metrics.

Would be trivial to hash so that normal aggregation worked, and shouldn't add a ton of overhead storage-wise.  Biggest issue would likely be the performance of sending the tags with the metric, just in terms of memory being moved around.

We might be able to do some `Scope`-esque magic with keeping a registry or something, but who knows.

tobz:
This PR accomplishes two major things:
- we've renamed the crates so that `metrics` is the facade crate, and `metrics-runtime` is the actual implementation
- we've added label support

Label support is partly due to changing the metric name to a composite `Key` -- thanks to @kazimuth for bringing this up in #23 -- as well as adding support for it to the facade crate and the `Sink` interface.

To-do list:
- [x] use labels in existing recorders
- [x] ensure all crates have `#![deny(missing_docs)]`
- [x] ensure formatting
- [x] bump the minor version for all of the renamed crates and for metrics-core (and update dependencies)
- [x] add documentation for label support in macros and in `Sink`

Note: we need to make sure that the versions we bump to are the _actual_ next minor version based on crates.io, and not the version of the crate when it had its old name.

Fixes #5, #23, #24.

/cc @kazimuth @LucioFranco


metrics-core/src/lib.rs
```diff
@@ -32,24 +32,172 @@
...
+}
+
+/// A metric key.
+///
+/// A key always includes a name, but can optional include multiple labels used to further describe
+/// the metric.
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct Key {
+    #[doc(hidden)]
+    pub name: ScopedString,
```
LucioFranco: why hidden public?

metrics-core/src/lib.rs
```diff
@@ -32,24 +32,172 @@
...
+        L: Into<Vec<Label>>,
+    {
+        Key {
+            name: name.into(),
+            labels: Some(labels.into()),
+        }
+    }
+
+    /// Name of this key.
+    pub fn name(&self) -> ScopedString {
```
LucioFranco: prob should return &ScopedString?

metrics-core/src/lib.rs
```diff
@@ -32,24 +32,172 @@
...
+                    .map(|label| format!("{} = {}", label.0, label.1))
+                    .collect::<Vec<_>>();
+                write!(f, "Key({}, [{}])", self.name, kv_pairs.join(", "))
+            }
+        }
+    }
+}
+
+/// A value that can be converted into a `Key`.
+pub trait IntoKey {
```
LucioFranco: isnt his just `Into<Key>`?

tobz: Yes, but we can't implement that for string types since we don't own them, and thus the custom trait.

metrics-core/src/lib.rs
tobz: Good call; I created the type originally thinking I'd use the fields, and then I decided on accessors, but never cleaned up the fields.

metrics-core/src/lib.rs
```diff
@@ -32,24 +32,206 @@
...
+
+/// A metric key.
+///
+/// A key always includes a name, but can optional include multiple labels used to further describe
+/// the metric.
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct Key {
+    name: ScopedString,
+    #[doc(hidden)]
+    labels: Option<Vec<Label>>,
```
samanpa: Consider using a plain Vec<Label> instead of Option<Vec<Label>>. The Vec::new is also very fast and there it is still easy to check if there are no labels.

```asm
playground::Option::new:
	movq	%rdi, %rax
	movq	$0, (%rdi)
	retq

playground::NoOption::new:
	movq	%rdi, %rax
	movq	$8, (%rdi)
	xorps	%xmm0, %xmm0
	movups	%xmm0, 8(%rdi)
	retq
```

tobz: Ah yeah, nice.  It actually looks like `Vec:new` doesn't allocate, so performance should be identical! :D

metrics-runtime/src/sink.rs
```diff
@@ -0,0 +1,640 @@
...
+/// Handle for sending metric samples.
+#[derive(Debug)]
+pub struct Sink {
+    metric_registry: Arc<MetricRegistry>,
+    metric_cache: FastHashMap<Identifier, ValueHandle>,
+    scope_registry: Arc<ScopeRegistry>,
+    scope: Scope,
+    scope_handle: ScopeHandle,
+    clock: Clock,
+    default_labels: Option<Vec<Label>>,
```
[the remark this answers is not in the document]
tobz: Yeah, good call.  I'll do that.

metrics-core/src/lib.rs
jean-airoldie: Why not implement `From<String> for Key` which automatically provides a `Into<Key> for String` via the [`blanket implementation`]?

[`blanket implementation`]: https://doc.rust-lang.org/std/convert/trait.Into.html#method.into

metrics-recorder-text/src/lib.rs
```diff
@@ -191,7 +194,13 @@ impl MetricsTree {
 
 impl Into<String> for TextRecorder {
```
jean-airoldie: Why not implement `From<TextRecorder>` for String? This also provides `Into<String> for TextRecorder`.

tobz: Ah yeah, good point.  Will change.

metrics-recorder-prometheus/src/lib.rs
```diff
@@ -29,80 +32,121 @@ impl PrometheusRecorder {
...
     fn clone(&self) -> Self {
         Self {
             output: get_prom_expo_header(),
+            histos: HashMap::new(),
             quantiles: self.quantiles.clone(),
         }
     }
 }
 
 impl Into<String> for PrometheusRecorder {
```
jean-airoldie: Same comment as before `From<PrometheusRecorder>` would be better.

metrics-runtime/src/common.rs
```diff
@@ -1,58 +1,68 @@
...
     Root,
 
     /// A nested scope, with arbitrarily deep nesting.
     Nested(Vec<String>),
 }
 
-impl MetricScope {
-    pub(crate) fn into_scoped(self, name: MetricName) -> String {
+impl Scope {
+    pub(crate) fn into_scoped(self, name: ScopedString) -> ScopedString {
```
jean-airoldie: The naming doesnt really follow the rust conventions (https://rust-lang-nursery.github.io/api-guidelines/naming.html#c-conv). I think you would want a special consuming constructor for scoped string instead.

metrics-core/src/lib.rs
tobz: Yeah, I think that'd work.  Apparently I forgot all about the `From` -> `Into` blanket impl when I wrote all of this... 😅

metrics-runtime/src/sink.rs
```diff
@@ -0,0 +1,640 @@
...
+    registry::{MetricRegistry, ScopeRegistry},
+};
+use fxhash::FxBuildHasher;
+use metrics_core::{IntoKey, IntoLabels, Key, Label, ScopedString};
+use quanta::Clock;
+use std::error::Error;
+use std::fmt;
+use std::sync::Arc;
+
+type FastHashMap<K, V> = hashbrown::HashMap<K, V, FxBuildHasher>;
```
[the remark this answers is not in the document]
tobz: `hashbrown` is becoming the libstd implementation in 1.36, which hasn't been released yet.  Can address it in a future PR once it's been out for long enough.

metrics-core/src/lib.rs
```diff
@@ -111,8 +296,11 @@ pub trait SnapshotProvider {
 
 /// A value that can provide on-demand snapshots asynchronously.
 pub trait AsyncSnapshotProvider {
+    /// Snapshot given by the provider.
     type Snapshot: Snapshot;
+    /// Errors produced during generation.
     type SnapshotError;
+    /// The future response value.
     type SnapshotFuture: Future<Item = Self::Snapshot, Error = Self::SnapshotError>;
```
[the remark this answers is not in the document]
tobz: Like we talked about, definitely soon. :)

metrics-recorder-prometheus/src/lib.rs
tobz: Will do.

metrics-core/src/lib.rs
tobz: Done!

metrics-runtime/src/common.rs
tobz: The problem is that `ScopedString` is actually just a type alias for `Cow<'static, str>`, so the best we could do, I think, is an extension trait, which feels like a lot of extra code.

Any thoughts?

metrics-core/src/lib.rs
```diff
@@ -32,24 +32,207 @@
...
+            name: name.into(),
+            labels: Vec::new(),
+        }
+    }
+
+    /// Creates a `Key` from a name and vector of `Label`s.
+    pub fn from_name_and_labels<N, L>(name: N, labels: L) -> Self
+    where
+        N: Into<ScopedString>,
+        L: Into<Vec<Label>>,
```
jean-airoldie: Seems unlikely that it would be usefull? Maybe `IntoIterator<Item = Label>`, followed by a collect would be more usefull here. Can't think of anything that implements `Into<Vec<T>>` from the top of my head.

metrics-core/src/lib.rs
```diff
@@ -32,24 +32,207 @@
...
+        Key {
+            name: name.into(),
+            labels: labels.into(),
+        }
+    }
+
+    /// Adds a new set of labels to this key.
+    ///
+    /// New labels will be appended to any existing labels.
+    pub fn add_labels(&mut self, new_labels: Vec<Label>) {
```
jean-airoldie: I think for consistency's sake you would want `new_labels` to be a generic like the one used in the `from_name_and_labels` constructor.

metrics-core/src/lib.rs
```diff
@@ -74,7 +257,7 @@ pub trait Recorder {
     /// counters and gauges usually have slightly different modes of operation.
     ///
     /// For the sake of flexibility on the exporter side, both are provided.
-    fn record_counter<K: Into<Key>>(&mut self, key: K, value: u64);
+    fn record_counter(&mut self, key: Key, value: u64);
```
jean-airoldie: I imagine that you removed the generic here so that `Recorder` can potentially be used as a trait object?

metrics-recorder-prometheus/src/lib.rs
```diff
@@ -1,12 +1,15 @@
 //! Records metrics in the Prometheus exposition format.
+#![deny(missing_docs)]
 use hdrhistogram::Histogram;
 use metrics_core::{Key, Recorder};
 use metrics_util::{parse_quantiles, Quantile};
+use std::collections::HashMap;
```
jean-airoldie: I see that you already have `hashbrown` as a dependency for you runtime. Why not use it everywhere since its basically a faster implementation?

metrics-core/src/lib.rs
```diff
@@ -32,24 +32,195 @@
...
+    L: IntoLabels,
+{
+    fn from(parts: (K, L)) -> Key {
+        let labels = parts.1.into_labels();
+        Key::from_name_and_labels(parts.0, labels)
+    }
+}
+
+/// A value that can be converted to `Label`s.
+pub trait IntoLabels {
```
jean-airoldie: Whats the advantage of this trait vs. using something like:
```rust
I: IntoIterator<Item = E>,
E: Into<Label>,
```
Then you could do something like:
```rust
impl From<(K, V)> for Label
where
  K: Into<ScopedString>,
  V: Into<ScopedString>,
````
So that you can convert an iterator of `(K, V)` into `Vec<Label>` like you do later.

File: metrics-recorder-prometheus/src/lib.rs [hunk 16]
```diff
@@ -29,80 +32,121 @@ impl PrometheusRecorder {
...
-        self.output.push_str(values.len().to_string().as_str());
-        self.output.push_str("\n");
     }
 }
 
 impl Clone for PrometheusRecorder {
     fn clone(&self) -> Self {
         Self {
             output: get_prom_expo_header(),
+            histos: HashMap::new(),
```
jean-airoldie: Seems like an unexpected property for a clone impl? I think the average user would assume that everything in the recorder would get duplicated. I think you would probably want to use a differently named method.

metrics-runtime/src/common.rs
```diff
@@ -1,58 +1,68 @@
...
 
-/// Optimized metric name.
-///
-/// This can either be a [`&'static str`](str) or [`String`].
-pub type MetricName = Cow<'static, str>;
-
 /// A scope, or context, for a metric.
 ///
 /// Not interacted with directly by end users, and only exposed due to a lack of trait method
 /// visbility controls.
```
jean-airoldie: If you would prefer to hide this to users i suggest you use `#[doc(hidden)]` which will remove it from the documentation.

metrics-runtime/src/common.rs
jean-airoldie: Right, I missed that. I guess you could to a weird conversion like `From<(Scope, String)> for ScopedString`. Or maybe  use another method name, although I can't think of anything right now.

metrics-runtime/src/data/snapshot.rs
```diff
@@ -1,31 +1,38 @@
...
 /// A point-in-time view of metric data.
 #[derive(Default, Debug)]
 pub struct Snapshot {
-    measurements: Vec<(String, ValueSnapshot)>,
+    measurements: Vec<(Key, ValueSnapshot)>,
 }
 
 impl Snapshot {
-    pub(crate) fn from(from: Vec<(String, ValueSnapshot)>) -> Self {
+    pub(crate) fn from(from: Vec<(Key, ValueSnapshot)>) -> Self {
```
jean-airoldie: I think it would probably better to use `From<Vec<(Key, ValueSnapshot)>>` here but hide the implementation using `#[doc(hidden)]`.

metrics-runtime/src/registry/metric.rs
```diff
@@ -25,19 +26,15 @@ impl MetricRegistry {
         }
     }
 
-    pub fn get_value_handle(&self, identifier: MetricIdentifier) -> MetricValue {
+    pub fn get_value_handle(&self, id: Identifier) -> ValueHandle {
```
jean-airoldie: The rust convention is to remove the `get_` entirely. So you would want `value_handle` here as your method name.

metrics-core/src/lib.rs
tobz: This is related to the key flexibility we want to provide.  I've been tweaking this stuff a lot so my explanation here could actually be wrong and it might be possible, but:

- we can't implement traits for a foreign type -- `impl Into<...> for T where T: Iterator`, etc -- so we added `IntoLabels` and then derive all the things that can fulfill it, which would be an existing vector of `Label`s or a referenced to an iterator of raw label pairs -- `(K, V) where K/V: Into<ScopedString>`.

This gives us the ability to derive an `Into<Key>` for `(N, L)` where `N` is the metric name and `L` fulfills one of the `IntoLabels` implementations.  Technically, since `Label` is public, we could just focus the constructor methods on handling the generic implementations, but to be honest, I just didn't really think of it.

Without sketching it out, I'm not sure if one is particularly better than the other?

I think with your idea, at least, I could make the `IntoLabels for &'a T` case cleaner by just requiring the item be to `Into<Label>` instead of handling the constraints directly.

Thoughts?

metrics-core/src/lib.rs
tobz: I originally did it that way because of the facade crate, yeah, wanting to be able to do `&'static Recorder`.

I haven't switched it over yet because I'm not sure it makes sense to force implementations to take multiple values at a time for histograms (`record_histogram(&mut self, key: Key, values: &[u64])`) since that was a design choice to help _exporters_ and code actually _recording_ metrics typically is sending a single value at a time.

metrics-recorder-prometheus/src/lib.rs
tobz: Yeah, that seems fine to do.  I'll get that done.

metrics-recorder-prometheus/src/lib.rs
tobz: So, the `Clone` implementation here is related to #11 and the design of how recorders/exporters work together.

Basically, exporters are long-lived, and use a recorder to convert each `Snapshot` they take into the intended output format.  The idea was that you would create a "master copy" of a recorder with all of the configuration it needed, like display options, formatting options, whatever you want, and it would clone it every time it took a snapshot.

So in this case, the `Clone` implementation is specialized so that it keeps the configuration data -- the quantiles for displaying histograms -- but doesn't keep the actual rendered metric data.

From #11, my gut feeling is that a builder pattern is the right thing to do, I just haven't done it yet.

Thoughts?

metrics-runtime/src/common.rs
tobz: Ah yeah, derp! I'll add that. :)

metrics-recorder-prometheus/src/lib.rs
tobz: All set.

metrics-runtime/src/common.rs
tobz: All set.

metrics-runtime/src/data/snapshot.rs
tobz: Yeah, I'm fine with generalizing the source data, but why the `#[doc(hidden)]`?  Crate visibility already lets us achieve that.

metrics-runtime/src/registry/metric.rs
tobz: I feel like it would actually be the other way around?  If you look at `Vec` or `HashMap`, they both have getters called `get`.

tobz: Or, alternative, maybe `get_or_register`, since it's technically do more of a HashMap entry API sort of operation than purely a get?

metrics-core/src/lib.rs
tobz: Yeah, that's a good point.  Will do.

metrics-runtime/src/data/snapshot.rs
jean-airoldie: Since your from impl is `pub(crate)` and `Sanpshot` is public, I assumed that you waited to hide it from users. And since `From<..>` is public you would need to hide it from the doc. But disregard that if that was not your intend.

metrics-runtime/src/registry/metric.rs
jean-airoldie: From what I understand, your only use `get` when your data only has one obvious thing that your could get. In the case of a `Vec`, that would be the data stored in the vec. I'll find a concrete example for the std lib.

jean-airoldie: But I get what you are saying. Using `value_handle` would give the false sense that the operation is inexpensive.

metrics-recorder-prometheus/src/lib.rs
jean-airoldie: Yeah that really sounds like a mutable builder pattern. Because currently the fact that you don't actually use the "master copy" directly for recording, your instead use a clone "shallow copy" is quite confusing. I feel that by having a separate `RecorderBuilder` and `Recorder` trait you make that distinction clearer.

Other than that, what is the rationale of having short lived recorders? It seems like your would have performance benefits from having a recorder that can be reset after each snapshot, so that you don't have to allocate each time.

jean-airoldie: For instance, your log recorder expects the recorder to implement `Into<String>` but I'm pretty sure your `Into<String>` implementation would have to allocate.

metrics-core/src/lib.rs
jean-airoldie: I think the main advantage of using the `IntoIterator` approach is that it is more interoperable. For instance, a user that uses a non-std container, lets say `SmallVec<Label>`, would need to convert its data into a `Vec<Label>` first. You also get a lot of stuff for free since almost every std container implements `IntoIterator`, including a `HashMap` which conveniently implements `IntoIterator<Item=(K,V)>`.

metrics-runtime/src/registry/metric.rs
jean-airoldie: `get_or_register` sound pretty good because it sounds like something fishy is going on.

jean-airoldie: Or maybe something that refers to the fact that this is basically a lazy initialization?

metrics-core/src/lib.rs
tobz: Done.

metrics-core/src/lib.rs
tobz: Yeah, let me try and rework this.  I'm still not sure if I can get rid of `IntoLabels` but it should be clearer once I play around with the code.

metrics-recorder-prometheus/src/lib.rs
tobz: The principal reason for preferring an owned recorder when processing a snapshot is that sharing a single recorder get very dicey with asynchronous callers, if not entirely impossible.  

Given the direction things are going with async/await and futures in Rust, I wanted `metrics` to be ahead of curve.  With that said, I'm not sure there's a place for a long-lived recorder that gets used over and over.

I'm definitely all for switching to a builder pattern, but I think I'd need to see a working example of a long-lived recorder being used from an asynchronous context (for example, using `metrics-exporter-http`) because I'm not sure it's actually possible.

jean-airoldie: Gotcha. This would be definitely possible using a lower level polling library, but I'm not sure its achievable with current tiokio / futures 0.1. The `Stream` API is quite a mess imo.

I guess the real question is whether calling `Clone` is expensive enough that it would be worth it.

metrics-runtime/src/registry/metric.rs
tobz: I changed this to `get_or_register`.

metrics-runtime/src/data/snapshot.rs
tobz: I ended up changing `from` to `new`, but leaving it `pub(crate)`.  Adding a proper `From` impl would have meant making `ValueSnapshot` public, which I didn't want to do.

metrics-core/src/lib.rs
tobz: I think I have a workable solution to this, but it's going to require some macro adjustments that involve more advanced macro syntax, so this is going to take me a little longer than just swapping the generic constraints. :)

metrics-core/src/lib.rs
tobz: Yeah, this is going to change once I redo the impls for getting the vector of labels.

metrics-core/src/lib.rs
tobz: I spent a lot of time trying to wrangle with the types, based on the common use cases, and here's why I'm deciding to keep `IntoLabels` after all:

The biggest issue came down to the fact that the there's a fundamental tension between callers who use macros (or the new `labels!` macro I added) which can generate a `Vec<Label>` directly and users who end up doing `&[("key", "value")]`, which has to generalize each tuple as `Into<Label>` or `K/V: Into<ScopedString>`.

If we had all of our methods take `IntoIterator<Item = Into<Label>>`, we'd have to unnecessarily allocate/copy a new vector even if we already had `Vec<Label>`, which is a performance penalty I'm not willing to pay.

Right now, the facade macros and the `labels!` helper macro can generate `Vec<Label>` for a user, which should over 99% of the use cases.  For the people that have a custom structure, there's `impl IntoLabels for &T where T: IntoIterator<Item = Into<Label>>` so they simply need to give us a reference to their container, and we'll do the copy for them.

It's not perfect, but I think it covers the intended use cases while still allowing flexibility.

jean-airoldie: If `IntoIterator<Item = Into<Label>>` does allocate like you claim, then this is the right decision. However I'm not sure it actually does. I'll make sure.

jean-airoldie: After looking at the generated assembly for `rustc 1.35`, I can confirm that it allocates. So I agree with your decision.

metrics-runtime/src/common.rs
```diff
@@ -1,58 +1,64 @@
...
-#[derive(PartialEq, Eq, Hash, Clone)]
-pub enum MetricScope {
+#[doc(hidden)]
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub enum Scope {
     /// Root scope.
     Root,
 
     /// A nested scope, with arbitrarily deep nesting.
     Nested(Vec<String>),
```
samanpa: Why not used a `Vec<ScopedString>`

PR #30, opened 2019-07-06
core: add Builder trait for recorders

### LINKED ISSUE #11: Is the recorder cloning approach really the best we can do?

tobz:
This change adds a new trait -- `Builder` -- which defines a value that can create new recorder instances.

As we have a need to generate owned recorders, in particular for futures-based code, the `Builder` trait provides a way to do so without any jankiness, such as the prior Clone-based approach.

As such, all exporters now expect a builder to be passed in, rather than the recorder itself.

Closes #11.

### DIFF
--- a/metrics-core/src/lib.rs
+++ b/metrics-core/src/lib.rs
@@ -284,6 +284,23 @@ pub trait Recorder {
     fn record_histogram(&mut self, key: Key, values: &[u64]);
 }
 
+/// A value that can build a recorder.
+///
+/// Recorders are intended to be single-use containers for rendering a snapshot in a particular
+/// format. As many systems are multi-threaded, we can't easily share a single recorder amongst
+/// multiple threads, and so we create a recorder per snapshot, tying them together.
+///
+/// A builder allows us to generate a recorder on demand, giving each specific recorder an
+/// interface by which they can do any necessary configuration, initialization, etc of the recorder
+/// before handing it over to the exporter.
+pub trait Builder {
+    /// The recorder created by this builder.
+    type Output: Recorder;
+
+    /// Creates a new recorder.
+    fn build(&self) -> Self::Output;
+}
+
 /// A value that holds a point-in-time view of collected metrics.
 pub trait Snapshot {
     /// Records the snapshot to the given recorder.
--- a/metrics-exporter-http/src/lib.rs
+++ b/metrics-exporter-http/src/lib.rs
@@ -16,31 +16,33 @@ use hyper::rt::run as hyper_run;
 use hyper::rt::Future;
 use hyper::service::service_fn;
 use hyper::{Body, Response, Server};
-use metrics_core::{AsyncSnapshotProvider, Recorder, Snapshot};
+use metrics_core::{AsyncSnapshotProvider, Builder, Snapshot};
 use std::error::Error;
 use std::net::SocketAddr;
+use std::sync::Arc;
 
 /// Exports metrics over HTTP.
-pub struct HttpExporter<C, R> {
+pub struct HttpExporter<C, B> {
     controller: C,
-    recorder: R,
+    builder: B,
     address: SocketAddr,
 }
 
-impl<C, R> HttpExporter<C, R>
+impl<C, B> HttpExporter<C, B>
 where
-    C: AsyncSnapshotProvider + Clone + Send + 'static,
-    C::SnapshotFuture: Send + 'static,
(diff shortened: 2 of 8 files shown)


tobz:
cc @jean-airoldie for thoughts

metrics-core/src/lib.rs
```diff
@@ -284,6 +284,23 @@ pub trait Recorder {
     fn record_histogram(&mut self, key: Key, values: &[u64]);
 }
 
+/// A value that can build a recorder.
+///
+/// Recorders are intended to be single-use containers for rendering a snapshot in a particular
+/// format.  As many systems are multi-threaded, we can't easily share a single recorder amongst
```
jean-airoldie: Whitespace typo

metrics-core/src/lib.rs
```diff
@@ -284,6 +284,23 @@ pub trait Recorder {
...
+///
+/// A builder allows us to generate a recorder on demand, giving each specific recorder an
+/// interface by which they can do any necessary configuration, initialization, etc of the recorder
+/// before handing it over to the exporter.
+pub trait Builder {
+    /// The recorder created by this builder.
+    type Output: Recorder;
+
+    /// Creates a new recorder.
+    fn create(&self) -> Self::Output;
```
jean-airoldie: Maybe `fn build(&self) -> Self::Output`? Seems more intuitive.

metrics-core/src/lib.rs
```diff
@@ -284,6 +284,23 @@ pub trait Recorder {
...
+/// A value that can build a recorder.
+///
+/// Recorders are intended to be single-use containers for rendering a snapshot in a particular
+/// format.  As many systems are multi-threaded, we can't easily share a single recorder amongst
+/// multiple threads, and so we create a recorder per snapshot, tying them together.
+///
+/// A builder allows us to generate a recorder on demand, giving each specific recorder an
+/// interface by which they can do any necessary configuration, initialization, etc of the recorder
+/// before handing it over to the exporter.
+pub trait Builder {
```
jean-airoldie: I'm wondering if this is the only builder trait that will ever be used by the library. If it is not, the naming would be problematic.

jean-airoldie:
I think this is way clearer and more intuitive that the prior Clone-based approach.

However I still don't like the concept of the single use `Recorder` as it forces unneeded allocations. I think I'll investigate it on my own to see if it is indeed required by the futures 0.1 runtime.

metrics-core/src/lib.rs
tobz: I'm pretty sure this is the only builder we'll need/want in `metrics-core` specifically, yeah.

metrics-core/src/lib.rs
tobz: Fixed.

metrics-core/src/lib.rs
tobz: Changed.

Issue #33, opened 2019-07-07
Moving to a reusable Recorder


jean-airoldie:
Related to #11, #30.

Currently, each time a snapshot is exported there are multiple (arguably) unnecessary allocations.

Lets say we look at the http metrics exporter code. This a small snippet extracted from https://github.com/metrics-rs/metrics/blob/master/metrics-exporter-http/src/lib.rs#L70.
```rust
// This is the main logic for the async event loop for the http exporter.
service_fn(move |_| {
    let recorder3 = recorder2.clone();

    controller2
        .get_snapshot_async()
        .then(move |result| match result {
            Ok(snapshot) => {
                let mut r = recorder3.clone();
                snapshot.record(&mut r);
                let output = r.into();
                Ok(Response::new(Body::from(output)))
            }
            Err(e) => Err(e),
        })
})
```
The following is some kind of pseudo code representation meant to illustrate the allocations made by this logic.

```
Controller::get_snapshot_async() {
    MetricsRegistry::get_snapshot() {
        Vec::new(); // allocates to hold metrics.
        ValueHandle::snapshot() {
            for value in values {
                Counter -> (),
                Gauge -> (),
                Histogram -> AtomicWindowedHistogram::snapshot() {
                    Vec::new(); // allocates to hold values
                }
            }
        }
    }
}
Recorder::clone() {
    HashMap::new();
    Vec::new();
}
Recorder::into<String>() {
  String::new();
}
```

So each time we allocate a new `Snapshot` and `Recorder` which is quite expensive.
What I suggest is that we get rid of the `Snapshot` entity and move to a reusable `Recorder`.
The logic would become something like this:
```rust
service_fn(move |_| {
    // The recorder should keep track of its state internally, so that it can
    // only be written to once.
    recorder.reset();
    // Takes a snapshot of the metrics and write them in the recorder. This might
    // allocate memory if needed.
    match controller.snapshot(&mut recorder) {
        Ok(_) => {
            let output: String = recorder.export();
            Ok(Response::new(Body::from(output)))
        }
        Err(e) => Err(e),
    }
})
```
This way we could reuse the allocated memory from the `Recorder` and only allocate the output `String`.

Admittedly, I am not currently certain that this is actually possible with the futures 0.1 runtime, but I'll work on a PR.

tobz:
I think it's definitely possible.

What I'd like to do is keep around the ability to take an independent snapshot, though, because it provides flexibility if users need to do something with the raw data.

[...]

PR #35, opened 2019-07-08
Recorder trait rework

jean-airoldie:
The objective of this PR is to allow for reusable recorders #33 and attempt to simplify the API.


metrics-recorder-text/src/lib.rs
```diff
@@ -159,15 +177,15 @@ impl MetricsTree {
         }
     }
 
-    pub fn into_output(self) -> String {
+    pub fn export_string(&self) -> String {
```
tobz: Maybe just call this `render`?

metrics-exporter-log/src/lib.rs
```diff
@@ -57,46 +55,26 @@ where
...
-        }
+        self.controller.record(&mut self.recorder);
+        let output = self.recorder.export();
+        log!(self.level, "{}", output);
+        self.recorder.flush();
     }
 
     /// Converts this exporter into a future which logs output on the given interval.
-    pub fn into_future(self, interval: Duration) -> impl Future<Item = (), Error = ()>
+    pub fn run_async(mut self, interval: Duration) -> impl Future<Item = (), Error = ()>
```
tobz: I don't think think this should be renamed.  It's not actually running anything for the user, it's giving them the future that _they_ need to run.

jean-airoldie: From what in understand of the [`convention`] , a `into_` convertion can't have any parameters. Thats why I renamed it. But I agree my new name sucks.

[`convention`]: https://rust-lang-nursery.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

jean-airoldie: For instance https://docs.rs/futures/0.1.28/futures/future/trait.IntoFuture.html.

LucioFranco: How about `interval_async`?

metrics-recorder-text/src/lib.rs
```diff
@@ -192,47 +210,47 @@ impl MetricsTree {
...
-        structure.into_output()
+        structure.render()
     }
 }
 
-enum SortEntry {
-    Inline(String),
-    Nested(String, MetricsTree),
+enum SortEntry<'a> {
+    Inline(&'a String),
```
LucioFranco: by chance would a `str` be better here?

metrics-recorder-text/src/lib.rs
```diff
@@ -192,47 +210,47 @@ impl MetricsTree {
...
+    fn name(&self) -> &'a String {
         match self {
             SortEntry::Inline(s) => s,
             SortEntry::Nested(s, _) => s,
         }
     }
 }
 
-impl PartialEq for SortEntry {
+impl<'a> PartialEq for SortEntry<'a> {
```
LucioFranco: I think you can elide this lifetime `SortEntry<'_>`?

metrics-recorder-text/src/lib.rs
jean-airoldie: Yah, not sure what I was thinking.

metrics-recorder-text/src/lib.rs
jean-airoldie: damn i didn't know this was a thing. cool.

metrics-exporter-log/src/lib.rs
jean-airoldie: I'm thinking of putting the `Duration` directly directly into the type instead. I think its an acceptable tradeoff since its unlikely that a user would create a `LogExporter` and exclusively use `turn`.

jean-airoldie: That way we would get the same API than the http-exporter.

tobz: Yeah, I'm fine with making the interval be a field on the exporter. 👍

metrics-exporter-http/src/lib.rs
```diff
@@ -27,21 +26,29 @@ pub struct HttpExporter<C, R> {
...
+    pub address: SocketAddr,
+}
+
 impl<C, R> HttpExporter<C, R>
 where
-    C: AsyncSnapshotProvider + Clone + Send + 'static,
-    C::SnapshotFuture: Send + 'static,
-    C::SnapshotError: Error + Send + Sync + 'static,
-    R: Recorder + Clone + Into<String> + Send + 'static,
+    C: Record + Clone + Send + 'static,
```
samanpa: I don't think Clone is needed here.

metrics-exporter-http/src/lib.rs
```diff
@@ -27,21 +26,29 @@ pub struct HttpExporter<C, R> {
...
+    R: Recorder + Clone + Export<String> + Send + 'static,
```
samanpa: Don't think Clone is needed.

metrics-exporter-log/src/lib.rs
```diff
@@ -26,77 +25,70 @@ pub struct LogExporter<C, R> {
...
+    /// The configurable logging level.
+    pub level: Level,
+    /// The interval at which to log.
+    pub interval: Duration,
 }
 
 impl<C, R> LogExporter<C, R>
 where
-    R: Recorder + Clone + Into<String>,
+    R: Recorder + Flush + Clone + Export<String>,
```
samanpa: Don't think Clone is needed here.
I also expect Flush to imply Recorder.

metrics-exporter-http/src/lib.rs
jean-airoldie: In this case `Clone` actually means duplicate the configuration which seems to be needed for futures. This is being addressed by #30 since the `Clone` trait is confusing.

metrics-core/src/lib.rs
```diff
@@ -284,34 +290,16 @@ pub trait Recorder {
...
-    type Snapshot: Snapshot;
-    /// Errors produced during generation.
-    type SnapshotError;
-    /// The future response value.
-    type SnapshotFuture: Future<Item = Self::Snapshot, Error = Self::SnapshotError>;
-
-    /// Gets a snapshot asynchronously.
-    fn get_snapshot_async(&self) -> Self::SnapshotFuture;
+/// A value that can record a view of its collected metrics to a recorder `Recorder`.
+pub trait Record {
```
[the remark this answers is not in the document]
jean-airoldie: I kinda agree. The thing is, there is a snapshot data type which is a high definition compressed snapshot. So we would have to rename that as well and I can't think of anything.

metrics-exporter-log/src/lib.rs
jean-airoldie: Right. Also I'm thinking of renaming `Flush` for something like `FlushableRecorder` so that its obvious. Its a bit long though.

metrics-core/src/lib.rs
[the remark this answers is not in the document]
jean-airoldie: The `Record` trait is more like, "allow a `Recorder` to take a peek at the collected metrics". It is to the discretion of the `Recorder` whether anything is recorded. So what about naming it something like `Measure` or `Observe`?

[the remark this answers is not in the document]
tobz: I'd be fine with `Observe`.  I think that fits.

metrics-core/src/lib.rs
```diff
@@ -255,63 +254,48 @@ impl AsNanoseconds for Duration {
...
-    type Snapshot: Snapshot;
-    /// Errors produced during generation.
-    type SnapshotError;
-    /// The future response value.
-    type SnapshotFuture: Future<Item = Self::Snapshot, Error = Self::SnapshotError>;
-
-    /// Gets a snapshot asynchronously.
-    fn get_snapshot_async(&self) -> Self::SnapshotFuture;
+/// A value whose metrics can be observed by an `Observer`.
+pub trait Observe {
```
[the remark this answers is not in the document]
tobz: I think this is just a language thing.  I would describe it as: anything that implements `Observe` can be _observed.  It's not emitting to us, we're pulling (observing) it.

I think this fits better with common terminology around the visitor/observed pattern.

metrics-core/src/lib.rs
```diff
@@ -255,63 +254,48 @@ impl AsNanoseconds for Duration {
     }
 }
 
-/// A value that records metrics.
-pub trait Recorder {
+/// A value that can export its content as a `T`.
```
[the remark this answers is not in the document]
tobz: I think @jean-airoldie just missed removing this line.

metrics-core/src/lib.rs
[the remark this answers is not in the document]
samanpa: Maybe Observable then to borrow from rx.net.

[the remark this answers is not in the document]
tobz: The general naming pattern for traits is verb form i.e. `Read`, `Write`, `Drain`, etc.

metrics-core/src/lib.rs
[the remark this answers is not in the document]
jean-airoldie: Indeed I did.

jean-airoldie:
Some thoughts:
* Should `Drain<T>` really imply `Observer`? If it does, should the trait be named so that it more obvious?
* Currently the `Snapshot` is implemented as a method on the controller. Should it be a trait instead?

tobz:
Yeah, the more I think about it, the more I like the idea of having a snapshot trait.  Just provides more modularity/flexibility for downstream users.

As far as `Drain<T>`, I don't think it should imply `Observe`, no.  I realize that without that, it's a very bare/simple trait, but from my perspective, that's fine.

(apologies for the slow reviews here, was on vacation)

jean-airoldie:
So the `LogExporter` now takes a `Builder` instead of an observer directly. This makes the syntax consistant with the `HttpExporter` and the extra cost is insignificant since the `Observer` is only built once.

jean-airoldie:
Also I'm unsure what to do with the `Snapshot`. Should it be a data structure? Another visitor trait? I don't really understand the purpose of the snapshot so I can't really reason about it.

tobz:
Eh, you know what? Fuck it.  Let's just leave `Controller::snapshot` for the time being.

Thinking about your comments makes me think we probably need to find a better use case to support having it be all trait-based, but I know that benchmarks/examples need the data structure as-is so leaving it as `Controller::snapshot` continues to provide that access while we figure out the long-term place for it.

I realize this is nit picky, but I still want to take another day to think through the multiple args vs arg struct change (i.e. `HttpExporterArgs`) because I'm not entirely sold on it.

jean-airoldie:
@tobz Yes I understand that it has bigger implications in terms of consistency. I think the arg struct thing might be a weird personal preference of mine. I'm not opposed using multiple args as long as they are properly documented on the function method.

I think the question is more about, how many positional arguments are too much in a function call. At some point the function becomes too confusing to use and an arg struct helps to alleviate that.

tobz:
Here's my vote: we switch back to individual arguments for now, and figure out if there's a better pattern we can come up with for composing observers/exporters, possibly with some sort of builder/fluent API.

Does that seem reasonable?

jean-airoldie:
Yes, that's what I was thinking about. Every exporter will need an observer and a composer so there is probably a decent intermediary abstraction that we could use instead.

tobz:
I'm gonna go ahead and merge this.  Everything looks pretty reasonable to me and we can do follow-ups for the things we talked about. 👍

PR #38, opened 2019-07-23
feature: add a JSON observer

### LINKED ISSUE #37: Create a JSON observer.
This is a follow-on from #10, where we want to add another observer: JSON.

This would have output similar to the text observer, but just in JSON format.  It'd maintain the hierarchy, etc.

tobz:
This adds a brand-new JSON observer to the mix.

We've refactored a bit, making the builder for the observers actually function like a builder, where methods to set the configuration update the builder fluently.

We've also extracted MetricsTree which is now serializable by serde. This means anything based on MetricsTree can implicitly be serialized by any serde serializer, which should make it easier for us to spin up new output formats in the future.

Closes #37

### DIFF
--- a/metrics-observer-json/Cargo.toml
+++ b/metrics-observer-json/Cargo.toml
@@ -1,19 +1,22 @@
 [package]
-name = "metrics-exporter-http"
-version = "0.1.2"
+name = "metrics-observer-json"
+version = "0.1.0"
 authors = ["Toby Lawrence <toby@nuclearfurnace.com>"]
 edition = "2018"
 
 license = "MIT"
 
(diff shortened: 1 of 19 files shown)


metrics-observer-json/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "metrics-observer-json"
+version = "0.1.0"
+authors = ["Toby Lawrence <toby@nuclearfurnace.com>"]
+edition = "2018"
+
```
LucioFranco: keywords?

tobz: Fixed

PR #39, opened 2019-07-24
runtime: add proxy metric support

### LINKED ISSUE #25: Lazy metrics evaluation
I'm currently using `rust-prometheus` in my applications for monitoring, but I'm interested in this crate since it decouples the actual metrics for the exporter. Also the performance overhead seems much lower. However I have a few questions regarding a specific use case I have.

I'm using prometheus to export process specific metrics such as cpu usage, disk usage, and network usage, provided by the OS. However, since actually getting of these metrics is relatively expensive, for instance parsing the content of the `/proc/<pid>/` directory, it must be done infrequently. Is it possible to have such lazy evaluated metrics that would only be evaluated when required? Or would I need to manually evaluate these metrics at a given time interval and push them?

tobz:
This adds support for a new internal metric type: `Proxy`.

This allows users to register a closure to run every time a snapshot is requested, or the metric state is observed.  This closure can return multiple measurements (counters, gauges, or histograms) that will be appended to the actual metrics register in the receiver.

This can be used for pulling in metrics that are otherwise hard or expensive to recalculate over and over, saving the CPU cycles and the extra integration points of recording them periodically.

Fixes #25.

/cc @jean-airoldie


metrics-runtime/src/common.rs
```diff
@@ -23,6 +25,17 @@ pub enum Scope {
 }
 
 impl Scope {
+    /// Adds a new part to this scope.
+    pub fn add_part(&mut self, part: ScopedString) {
```
tobz: I should probably switch `Sink` to use this new method instead of manually destructing the scope and adding things manually.

metrics-runtime/src/registry/metric.rs
```diff
@@ -65,34 +66,63 @@ impl MetricRegistry {
...
+                ValueSnapshot::Multiple(mut measurements) => {
+                    // Tack on the key name that this proxy was registered with to the scope so
+                    // that we can clone _that_, and then scope our individual measurements.
+                    let mut scope = scope.clone();
+                    let key_parts = key.into_parts();
+                    scope.add_part(key_parts.0);
+
+                    for (subkey, measurement) in measurements.drain(..) {
+                        let scope = scope.clone();
+                        let subkey = subkey.map_name(|name| scope.into_scoped(name));
```
tobz: We should probably accept labels for proxy creation and then apply them here to the subkey after splitting them from the proxy key.

metrics-runtime/src/common.rs
tobz: All set!

metrics-runtime/src/registry/metric.rs
tobz: This is also now done.

PR #46, opened 2019-08-13
Improved gauge, timing and value macro doc

jean-airoldie:
Further doc improvements.

### DIFF
--- a/metrics/src/macros.rs
+++ b/metrics/src/macros.rs
@@ -43,18 +43,22 @@ macro_rules! counter {
     };
 }
 
-/// Records a gauge.
+/// Update a gauge with a value.
+///
+/// This will register a gauge with the given name, if it does not already
+/// exist, then update it, replacing the previous value with given value. Optionally,
+/// a set of labels, of the form `key => value`, can be passed to further
+/// describe the gauge.
 ///
 /// Functionally equivalent to calling [`Recorder::record_gauge`].
 ///
 /// ### Examples
 ///
 /// ```rust
-/// # #[macro_use]
-/// # extern crate metrics;
+/// use metrics::gauge;
+///
 /// fn update_current_value() {
-///     let value: i64 = -131;
-///     gauge!("current_value", value);
+///     gauge!("current_value", -131);
 /// }
 /// # fn main() {}
 /// ```
@@ -62,8 +66,8 @@ macro_rules! counter {
 /// Labels can also be passed along:
 ///
 /// ```rust
-/// # #[macro_use]
-/// # extern crate metrics;
+/// use metrics::gauge;
+///
 /// fn update_current_value() {
 ///     let value: i64 = -131;
 ///     let creator: String = String::from("jane");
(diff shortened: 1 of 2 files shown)


metrics/src/macros.rs
```diff
@@ -122,8 +130,7 @@ macro_rules! gauge {
 /// Labels can also be passed along:
 ///
 /// ```rust
-/// # #[macro_use]
-/// # extern crate metrics;
+/// # use metrics::timing;
```
tobz: I think we should remove the `#` so that this line is shown?

metrics/src/macros.rs
```diff
@@ -86,13 +90,17 @@ macro_rules! gauge {
 
 /// Records a timing.
 ///
+/// This will register an histogram with the given name, if it does not already
+/// exist, then add data point with the given timing. This timing must implement
+/// `AsNanoseconds`. Optionally, a set of labels, of the form `key => value`,
```
tobz: You have to wrap the type in brackets -- ``[`AsNanoseconds`]`` -- for it to become a hyperlink.

metrics/src/macros.rs
```diff
@@ -172,6 +181,10 @@ macro_rules! timing {
 
 /// Records a value.
 ///
+/// This will register an histogram with the given name, if it does not already
+/// exist, then add data point with the given value. Optionally, a set of labels,
+/// of the form `key => value`, can be passed to further describe the counter.
```
tobz: `counter` -> `value`

metrics/src/macros.rs
```diff
@@ -86,13 +90,17 @@ macro_rules! gauge {
 
 /// Records a timing.
 ///
+/// This will register an histogram with the given name, if it does not already
+/// exist, then add data point with the given timing. This timing must implement
+/// `AsNanoseconds`. Optionally, a set of labels, of the form `key => value`,
+/// can be passed to further describe the counter.
```
tobz: `counter` -> `timing`

metrics/src/macros.rs
```diff
@@ -43,27 +43,31 @@ macro_rules! counter {
...
 }
 
-/// Records a gauge.
+/// Update a gauge by a value.
 ///
-/// Functionally equivalent to calling [`Recorder::record_gauge`].
+/// This will register a gauge with the given name, if it does not already
+/// exist, then update it by the given value. A positive value will increment it,
+/// while a negative value will decrement it. Optionally, a set of labels,
+/// of the form `key => value`, can be passed to further describe the counter.
```
tobz: `counter` -> `gauge`

metrics/src/macros.rs
jean-airoldie: a linting tool for dead links would be a godsend.

jean-airoldie:
After testing, I found out that I made a mistake in the doc. The gauge! macro indeed takes an absolute value, so ill update that.

PR #47, opened 2019-08-13
Renamed metrics::Recorder & metrics-runtime::Sink methods

jean-airoldie:
* Renamed record_counter to increment_counter
* Renamed record_gauge to update_gauge

### DIFF
--- a/metrics/src/lib.rs
+++ b/metrics/src/lib.rs
@@ -78,11 +78,11 @@
 //! struct LogRecorder;
 //!
 //! impl Recorder for LogRecorder {
-//!     fn record_counter(&self, key: Key, value: u64) {
+//!     fn increment_counter(&self, key: Key, value: u64) {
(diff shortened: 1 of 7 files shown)


metrics/src/lib.rs
```diff
@@ -324,12 +324,12 @@ pub fn recorder() -> &'static dyn Recorder {
 
 #[doc(hidden)]
 pub fn __private_api_record_count(key: Key, value: u64) {
```
tobz: `__private_api_record_count` -> `__private_api_increment_counter`?

jean-airoldie: missed that

tobz:
Should we also update the documentation for `gauge!`, `timing!`, and `value!`? The stuff you did in your last PR: changing the wording, fixing the macro import, etc.

jean-airoldie:
This breaks however the links in `metrics` doc, so Ill fix that.

PR #53, opened 2019-10-17
Fix tests and update some dependencies

bstrie:
Two methods were renamed in https://github.com/metrics-rs/metrics/commit/50db390ad7158b2f6a856f95cbab933af1bd19f3 , and this commit updates the doc comments to reflect that, as well as updating the test so that `cargo test` passes.

In addition, since Rust 1.36 the Rust standard library has included hashbrown as its backing implementation for std::collections::HashMap, so an explicit dependency involves redundant work for downstream users. If you wish to allow hashbrown to be used by users on toolchains preceding Rust 1.36, I can make hashbrown into a Cargo feature so that users can decide whether or not to include it.

This also bumps the parking_lot version from ^0.8 to ^0.9, for the entirely selfish reason that we'd like to begin using metrics for our project and I'm trying to avoid having multiple incompatible versions of dependencies in the crate graph. :P


tobz:
👋 👋 👋 

This is great, thank you!  It does seem like we're probably safe here re: hashbrown.  If someone is on <1.36 and they really need that extra performance, we can come to that road when they're about to cross it.

bstrie:
I was curious if you'd be interested in tagging a new release to crates.io? Of course, with the method renamings it would have to be a semver-incompatible release. I can file a PR bumping the version number if you'd like.

tobz:
Yeah, I'm working on getting the releases out shortly. 👍

tobz:
Alright, I released `metrics-runtime@0.12.0`.

The big jump from 0.1.0 to 0.12.0 was because I wanted to keep continuity with the package even after it was renamed from `metrics` to `metrics-runtime`.  I meant to do that with 0.11.0 but clearly got things jumbled up.

bstrie:
Awesome, thanks!

PR #55, opened 2019-11-11
Don't evaluate macro arguments with no recorder

TomGillen:
Modified the macros to not evaluate their arguments when no recorder is set. This avoids any `.to_string()` calls the user may be making for label values and avoids the `Vec` allocation for labels.

The benchmark, modified to convert a `usize` to a string for a label value, shows the following:

```
counter/no labels       time:   [2.5911 ns 2.6012 ns 2.6118 ns]
                        change: [-74.045% -73.926% -73.818%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) low mild
counter/with labels     time:   [1.7592 ns 1.7649 ns 1.7710 ns]
                        change: [-99.396% -99.384% -99.374%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
```

This brings the cost of using the macros when there is no recorder set close to nothing. I have been evaluating using `metrics` to instrument [legion](https://github.com/TomGillen/legion), and the not totally insignificant overhead of hitting the system allocator a few times on each metric update even in the (highly likely) case where the host application is not actually collecting them, is not ideal. I would be much more comfortable including metrics if there was negligable impact when they were at least not being used.

It would be even better if we could cache the counter references, like the `metrics-runtime` API allows, but that would be a much more involved change.

### DIFF
--- a/metrics/src/lib.rs
+++ b/metrics/src/lib.rs
@@ -303,27 +303,42 @@ impl error::Error for SetRecorderError {
 ///
 /// If a recorder has not been set, a no-op implementation is returned.
 pub fn recorder() -> &'static dyn Recorder {
+    static NOOP: NoopRecorder = NoopRecorder;
+    try_recorder().unwrap_or(&NOOP)
+}
+
+/// Returns a reference to the recorder.
+///
+/// If a recorder has not been set, returns `None`.
+pub fn try_recorder() -> Option<&'static dyn Recorder> {
     unsafe {
         if STATE.load(Ordering::SeqCst) != INITIALIZED {
-            static NOOP: NoopRecorder = NoopRecorder;
-            &NOOP
+            None
         } else {
-            RECORDER
+            Some(RECORDER)
         }
     }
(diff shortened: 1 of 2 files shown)


tobz:
Woohoo!

I'll admit, it's a little embarrassing that I never gave this much thought given the time I've put into `metrics-runtime` itself. 😬

I'm definitely veryyyy interested in a solution to cache the metric handles like `metrics-runtime` does.  Happy to hear any thoughts you might have.  I was thinking originally about cribbing from what `tracing` does with `Callsite` statics, but couldn't develop the idea in a way to also work with labels. 

At any rate, I love this PR, and appreciate it! I'll merge it now and can cut a release tomorrow.

tobz:
Work got crazy.  Released `metrics@0.12.1` including this.

Cheers!

PR #65, opened 2020-01-09
Migrate to core::futures

dfoxfranke:
This drops all dependencies on the 'futures' crate, and upgrades to
hyper-0.13 and tokio-0.2. Methods that return futures now use
async/await.

Since hyper has removed hyper_run(), I have in turn removed the
HttpExporter::run() method. The user should just call into_future()
and spawn or await the result however they wish.

### DIFF
--- a/metrics-exporter-http/src/lib.rs
+++ b/metrics-exporter-http/src/lib.rs
@@ -4,8 +4,6 @@
 //! via [`Drain<String>`].  It will respond to any requests, regardless of the method or path.
 //!
 //! # Run Modes
-//! - `run` can be used to block the current thread, running the HTTP server on the configured
-//! address
 //! - `into_future` will return a [`Future`] that when driven will run the HTTP server on the
 //! configured address
 #![deny(missing_docs)]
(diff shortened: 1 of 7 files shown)


metrics-exporter-http/src/lib.rs
```diff
@@ -45,33 +41,24 @@ where
...
         let builder = self.builder;
         let address = self.address;
 
-        build_hyper_server(controller, builder, address)
+        build_hyper_server(controller, builder, address).await
     }
 }
 
-fn build_hyper_server<C, B>(
+async fn build_hyper_server<C, B>(
```
tobz: We should collapse this into `into_future` because it's only used there now.

dfoxfranke: Okay, done. But now I notice the whole `HttpExporter` struct is looking a little silly, because calling `into_future()` is literally the only thing you can do with it. Perhaps it should just be replaced with an async fn that returns the future directly?

tobz: Yeah, let me think on this a little bit... I had always sort of intended this this crate would be equally usable in both sync and async code.  Gotta figure out what the right API looks like here.

dfoxfranke: I suppose I could add back the `run()` method, and have it set up a basic_scheduler tokio runtime and then block_on the future? Not sure how appealing that's going to be, I think most users' attitude is going to be that if their program is going to have a tokio runtime in it, they don't want it abstracted away.

tobz:
@dfoxfranke I made some small naming tweaks: something bugged me about the idea of an `into_*` method that is async since I prefer to see the be the verb of what it intends to do i.e. it should "run" an HTTP server, etc.

Long story short: `into_future` -> `async_run`, plus fixed the docs and facade example to match.  Gonna merge this once CI passes.

Thanks again for submitting this and unknowingly giving me a kick in the ass to get moving on this crate again. 🙇

PR #73, opened 2020-05-13
Add buckets (histogram) to Prometheus observer

jeromegn:
The API is awkward, I'm definitely open to suggestions!

Overriding buckets requires knowing the "rendered" name of the metric which is probably not  desirable.

```rust
let observer = metrics_observer_prometheus::PrometheusBuilder::new()
    .set_buckets(&[
        5_000_000,     // 5ms
        25_000_000,    // 25ms
        50_000_000,    // 50ms
        100_000_000,   // 100ms
        200_000_000,   // 200ms
        1_000_000_000, // 1s
    ])
    .set_buckets_for_metric(
        "my_sometimes_fast_metric",
        &[
            1_000_000,   // 1ms
            5_000_000,   // 5ms
            25_000_000,  // 25ms
            50_000_000,  // 50ms
            100_000_000, // 100ms
            200_000_000, // 200ms
        ],
    )
    .set_buckets_for_metric(
        "my_usually_fast_metric",
        &[
            1_000,     // 1us
            20_000,    // 20us
            100_000,   // 100us
            1_000_000, // 1ms
            5_000_000, // 5ms
        ],
    );
```

### DIFF
--- a/metrics-observer-prometheus/src/lib.rs
+++ b/metrics-observer-prometheus/src/lib.rs
@@ -3,6 +3,7 @@
 use hdrhistogram::Histogram;
 use metrics_core::{Builder, Drain, Key, Label, Observer};
 use metrics_util::{parse_quantiles, Quantile};
+use std::iter::FromIterator;
 use std::{collections::HashMap, time::SystemTime};
 
 /// Builder for [`PrometheusObserver`].
@@ -161,12 +162,23 @@ impl Drain<String> for PrometheusObserver {
                 output.push_str("\n");
             }
         }
+        let mut sorted_overrides = self
+            .buckets_by_name
+            .as_ref()
+            .map(|h| Vec::from_iter(h.iter()))
+            .unwrap_or_else(|| vec![]);
+        sorted_overrides.sort_by(|(a, _), (b, _)| b.len().cmp(&a.len()));
 
         for (name, mut by_labels) in self.histos.drain() {
-            let buckets = self
-                .buckets_by_name
-                .as_ref()
-                .and_then(|h| h.get(&name))
+            let buckets = sorted_overrides
+                .iter()
+                .find_map(|(k, buckets)| {
+                    if name.ends_with(*k) {
+                        Some(*buckets)
+                    } else {
+                        None
+                    }
+                })
                 .unwrap_or(&self.buckets);
             let use_quantiles = buckets.is_empty();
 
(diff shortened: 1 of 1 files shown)


metrics-observer-prometheus/src/lib.rs
```diff
@@ -138,21 +163,56 @@ impl Drain<String> for PrometheusObserver {
         }
 
         for (name, mut by_labels) in self.histos.drain() {
+            let buckets = self
+                .buckets_by_name
+                .as_ref()
+                .and_then(|h| h.get(&name))
+                .unwrap_or(&self.buckets);
+            let use_quantiles = buckets.is_empty();
+
```
tobz, line 172: This is really succinct; nice!

jeromegn:
This is now deployed in production for us. Appears to work as intended.

jeromegn:
@tobz are these checks suppose to pass? :D

tobz:
@jeromegn Something about the MSRV checks is messed up in Azure Pipelines and I haven't spent any time investigating it.

From looking at the diff, I don't see anything that could reasonably break the current target MSRV, so I think we can safely ignore their failure for this PR.

I promise I'll take a look at this tonight. 😭

metrics-observer-prometheus/src/lib.rs
```diff
@@ -28,6 +35,23 @@ impl PrometheusBuilder {
...
+    ///
+    /// Buckets values represent the higher bound of each buckets.
+    pub fn set_buckets(mut self, values: &[u64]) -> Self {
+        self.buckets = values.to_vec();
+        self
+    }
+
+    /// Sets the buckets for a specific metric, overidding the default.
+    ///
+    /// Matches the metric name using `ends_with`.
```
tobz, line 49: Can you extend this comment to mention that it not only matches the suffix, but searches for the longest match first?

metrics-observer-prometheus/src/lib.rs
```diff
@@ -28,6 +35,23 @@ impl PrometheusBuilder {
         self.quantiles = parse_quantiles(quantiles);
         self
     }
+
+    /// Sets the buckets to use when rendering summaries.
+    ///
+    /// Buckets values represent the higher bound of each buckets.
```
tobz, line 41: Can you add a sentence here (and to `set_buckets_for_metric` as well) that specifying buckets will change the default behavior from exporting a histogram and instead will export a summary for the given metric(s)?

jeromegn:
Pushed some comment changes, let me know if that's better :)

PR #79, opened 2020-07-18
Add atomic_shim to support platforms without 64-bit atomics

nazar554:
`atomic-shim` does target detection, so it is using `std::atomic` where possible.
See https://github.com/metrics-rs/quanta/issues/22 and https://github.com/metrics-rs/quanta/pull/23
There are no static `AtomicI64/AtomicU64` used so no `ctor` dependency is needed


tobz:
`metrics-runtime@0.13.1` is now released.  Let me know how that works for you. 👍

nazar554:
@tobz Nice, now I can build [vector](https://github.com/timberio/vector), just need to add the same fix there as well.

PR #80, opened 2020-07-20
refactor: next-generation metrics

tobz:
This PR represents the next generation of the `metrics` ecosystem, with a number of significant changes:

<list changes here when laptop battery isn't about to die>


gorzell:
> <list changes here when laptop battery isn't about to die>

It's too bad that this never happened.  Since there were also no release notes, it has been a real struggle to figure out how to upgrade to the new version.

tobz:
@gorzell There was certainly no upgrade guide, although the changelogs should be accurate in terms of what changed between versions.

Beyond that, the documentation for the crates -- i.e. https://docs.rs/metrics -- has been worked over pretty significantly and should hopefully be helpful in understanding how the pieces fit together.

The macros themselves didn't change much between using them with `metrics-runtime` and using them with the new exporter set-up, so I'd be interested to hear about the troubles you've been having in that regard.  Feel free to file issues for any of that stuff, or anything else really. We're also available in Discord -- https://discord.gg/eTwKyY9, `#metrics` -- to chat in real-time if that's more your speed.

gorzell:
Thanks for the pointer to the change logs.  For some reason I expected that to be at the top level of the repo, esp with crates being added/removed. I'll have a look through those see if there is anything else worth an issue or Discord chat.

Also, thanks for the quick reply.

PR #88, opened 2020-09-06
Support for using tracing span fields as metrics labels (dynamic labels remix)

tobz:
This is a remix of #87 where we solve the issue of cached callsites by pushing dynamism directly into `Label` itself, so that the macros can understand the full scope of when caching is safe to perform.

It accomplishes this by utilizing a resolved/unresolved paradigm, letting labels be declared upfront as "dynamic" so that downstream layers/recorders can potentially "resolve" them by providing a value, such as the tracing layer checking all current span fields and using those values.

An example:

    // A counter with entirely static labels.  "Static" here defines a label that is fixed and can't
    // be updated by a downstream layer, not specifically that both key/value are `&'static str`.
    counter!("my_counter", 1, "key" => "value");

    // A counter with a dynamic label.  This allows downstream layers to inspect it, and if desired,
    // update the value of the label with whatever they want.
    counter!("my_counter", 1, "dynamic_key");

Some design assumptions here:

All layers/recorders will be forced to filter out dynamic labels that are not yet resolved.  This is an ergonomics snag that we could potentially solve by adding another type that deals with entirely unresolved labels, and then have `Label` be a resolved label that always has a key/value.  Not sure how big of an issue this is, since once your recorder handles it, you're done.

This approach does not address the underlying issue with downstream layers/recorders _adding_ new labels to a key during registration.  One potential solution is to add the ability for recorders to declare that they're going to modify the key, or all keys, such that the callsite can detect whether or not it's safe to cache the identifier.  This might involve a tweaked identifier that exposes its own cacheability, or some new `Recorder` trait method that determines if a given `Key` would be altered if registered.


tobz:
Closing this as we merged #87

Issue #99, opened 2020-10-05
metrics-exporter-tcp: drop metrics when no clients are connected.


tobz:
(This is an issue for the yet-to-be-released `metrics-exporter-tcp`, an exporter that emits metrics over TCP to downstream clients using a Protocol Buffers-based payload.)

Currently, we always signal to the event loop that a new metric is available for sending, and go through the process of encoding them, enqueueing them, etc.  If we have no clients, there's no reason to do all of that work.

We should just be able to add in an atomic counter to track connected clients, and have the ingest codepath short-circuit if no clients are currently connected.  We might miss one or two metrics that way when a client first connects, but I don't think it's an issue, really, and it should drop the CPU usage when idle to almost nothing.

tobz:
Closed via #112.

PR #112, opened 2020-10-24
metrics-exporter-tcp: drop metrics when no clients are connected.

tobz:
This PR fixes the issue laid out in #99, which is that we go through with sending metrics to the internal event loop even no clients are connected, which slows down callers emitting metrics.

Granted, they'll end up paying the cost as soon as a single client connects, but we can at least reduce CPU usage where possible for optimal efficiency.

### DIFF
--- a/metrics-exporter-tcp/examples/tcp_client.rs
+++ b/metrics-exporter-tcp/examples/tcp_client.rs
@@ -25,9 +25,9 @@ fn main() {
             Err(e) => eprintln!("read error: {:?}", e),
         };
 
-        match proto::Metric::decode_length_delimited(&mut buf) {
+        match proto::Event::decode_length_delimited(&mut buf) {
             Err(e) => eprintln!("decode error: {:?}", e),
-            Ok(msg) => println!("metric: {:?}", msg),
+            Ok(msg) => println!("event: {:?}", msg),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


markjin1990:
I encountered a problem when using `metric-exporter-tcp` and [metrics-observer](https://github.com/metrics-rs/metrics/tree/main/metrics-observer) together that might be related to this feature. Not sure, if it is expected.

The problem is that I often lose some metrics (reported at the early stage in the function that collects metrics) on the client side, i.e., metric observer, if I do this
1. [me] Manually start metrics-observer listening on x.x.x.x:1234
2. [rust program] Start metric-exporter-tcp on x.x.x.x:1234
3. [rust program] Start running the function that collects metrics

A temporary workaround I am using that avoid losing metrics on metric observer,
1.  [rust program]Start metric-exporter-tcp on x.x.x.x:1234, and immediately sleep 10 seconds.
2. While [rust program] is sleeping, [me] manually start metrics-observer listening on x.x.x.x:1234
3. [rust program] After sleep is over, start running the function that collects metrics

tobz:
It's expected.

If we were able to send all metrics to any new client, it would mean holding all metrics seen forever, and then immediately sending them all to each new client.

However, I'd accept a PR to add that functionality behind an exporter configuration method or something.

PR #116, opened 2020-10-27
metrics-exporter-prometheus: split off tokio and hyper functionality to another (default) feature

kilpatty:
I threw this up relatively quickly, I can dive into it more tonight and improve documentation + ci (adding different features to actions).

What this PR does it make the Tokio and hyper functionality apart of a feature. It is enabled by default, but allows for users to replace disable it if they wish. The only major change is introducing another `build` function on the PrometheusBuilder that only returns the Recorder rather than both the Recorder and Exporter. This function is only enabled when the `tokio-exporter` feature is disabled.

### DIFF
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -15,14 +15,20 @@ readme = "README.md"
 categories = ["development-tools::debugging"]
 keywords = ["metrics", "telemetry", "prometheus"]
 
+[features]
+default = ["tokio-exporter"]
+tokio-exporter = ["hyper", "tokio"]
+
 [dependencies]
-metrics = { version = "0.13.0-alpha.1", path = "../metrics" }
-metrics-util = { version = "0.4.0-alpha.1", path = "../metrics-util"}
+metrics = { version = "0.13.0-alpha.1", path="../metrics" }
+metrics-util = { version = "0.4.0-alpha.1", path="../metrics-util" }
 hdrhistogram = "7.1"
-hyper = { version = "0.13", default-features = false, features = ["tcp"] }
-tokio = { version = "0.2", features = ["rt-core", "tcp", "time", "macros"] }
 parking_lot = "0.11"
 thiserror = "1.0"
+
+# Optional
+hyper = { version = "0.13", default-features = false, features = ["tcp"], optional = true }
+tokio = { version = "0.2", features = ["rt-core", "tcp", "time", "macros"], optional = true }
 
 [dev-dependencies]
 quanta = "0.6"
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -20,8 +20,8 @@ default = ["tokio-exporter"]
 tokio-exporter = ["hyper", "tokio"]
 
 [dependencies]
-metrics = { version = "0.13.0-alpha.1", path="../metrics" }
-metrics-util = { version = "0.4.0-alpha.1", path="../metrics-util" }
+metrics = { version = "0.13.0-alpha.1", path = "../metrics" }
+metrics-util = { version = "0.4.0-alpha.1", path = "../metrics-util" }
 hdrhistogram = "7.1"
 parking_lot = "0.11"
 thiserror = "1.0"
--- a/metrics-exporter-prometheus/src/lib.rs
+++ b/metrics-exporter-prometheus/src/lib.rs
@@ -334,7 +334,7 @@ pub struct PrometheusRecorder {
 }
 
 impl PrometheusRecorder {
-    ///Returns a [`PrometheusHandle`] from the inner struct of this recorder.
+    /// Gets a [`PrometheusHandle`] to this recorder.
     pub fn handle(&self) -> PrometheusHandle {
         PrometheusHandle {
             inner: self.inner.clone(),
@@ -351,7 +351,9 @@ impl PrometheusRecorder {
     }
 }
 
-/// A wrapper around the Inner struct of the Prometheus recorder.
+/// Handle to [`PrometheusRecorder`].
+///
+/// Useful for exposing a scrape endpoint on an existing HTTP/HTTPS server.
 pub struct PrometheusHandle {
     inner: Arc<Inner>,
 }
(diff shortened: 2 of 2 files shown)


metrics-exporter-prometheus/src/lib.rs
```diff
@@ -439,12 +449,31 @@ impl PrometheusBuilder {
...
+            buckets: self.buckets.clone(),
+            buckets_by_name: self.buckets_by_name,
+            descriptions: RwLock::new(HashMap::new()),
+        });
+
+        let recorder = PrometheusRecorder { inner };
+
+        Ok(recorder)
+    }
+
```
tobz, line 469: The problem here is that you actually need to install `PrometheusRecorder` via something like `metrics::set_boxed_recorder` (see line 429 in your version of the file) otherwise nothing is actually recording the metrics.

I think what I'd like to see overall here is `build`/`install` staying as-is, feature gated as you have them now, but two new variants -- not feature gated because they can exist even if `tokio`/`hyper` are being brought in -- called `build__with_handle`/`install_with_handle`. (I'm not super tied to the naming, but..)

What you would do here is create a new type, something like `PrometheusHandle`, which takes a clone of the inner state -- `Arc<Inner>` -- and instead of putting that new `render` function on `PrometheusRecorder`, you put it on `PrometheusHandle` instead.

Then, you'd emulate `build`/`install` but instead of having them spawn a thread or build a future, just pass back the handle.  This would better match what we have now, giving people the ability to install the recorder as-is or to get it back so they can wrap it in a higher-level construct like `Stack`/`Layer`... albeit with a new-found handle to access the data as they require.

Make sense?

metrics-exporter-prometheus/Cargo.toml
```diff
@@ -15,15 +15,21 @@ readme = "README.md"
...
 
+[features]
+default = ["tokio-exporter"]
+tokio-exporter = ["hyper", "tokio"]
+
 [dependencies]
-metrics = { version = "0.13.0-alpha.1", path = "../metrics" }
-metrics-util = { version = "0.4.0-alpha.1", path = "../metrics-util"}
+metrics = { version = "0.13.0-alpha.1", path="../metrics" }
+metrics-util = { version = "0.4.0-alpha.1", path="../metrics-util" }
```
tobz, line 24: Did you mean to scrub the spaces around the equal sign here?

tobz (changes requested):

metrics-exporter-prometheus/Cargo.toml
kilpatty: Sorry circling back to this now, will fix this -> Did not mean to scrub them, I think this was a result of me testing my previous issue of the versions not seeming to be updated.

metrics-exporter-prometheus/src/lib.rs
kilpatty: Makes sense, I was just looking at set_recorder which I believe takes a reference? So I was thinking we would still have access to the Recorder afterwards. 

I don't mind the build_with_handle/install_with_handle syntax, but just a thought what if instead of having separate functions we just had a helper function on the Recorder that returned the handle?

That way the process would look like this -> Call build, let handle = recorder.get_handle(), then call set_boxed_recorder. 

Just a thought, but I'm open to doing it either way.

kilpatty:
@tobz Pushed up an implementation of the handle function to return a PrometheusHandle which is simply a type alias around `Arc<Inner>`. Curious on your thoughts on this way of doing it. 

Also I can't promise my documentation comments are the best, but happy to have them ripped apart haha.

metrics-exporter-prometheus/src/lib.rs
tobz: Let me think about it for a little while.

One thing we definitely need is for `PrometheusHandle` to be its own type.  If we simply alias it over `Arc<Inner>`, we're now exposing other internal types, which is suboptimal.

kilpatty: I'll make that switch now

kilpatty: Pushed up fixes for type -> struct conversion, and eliminated the now not-needed doc comments on Inner

metrics-exporter-prometheus/src/lib.rs
```diff
@@ -439,12 +463,31 @@ impl PrometheusBuilder {
         Ok(())
     }
 
+    /// Builds the recorder and returns it.
+    /// This function is only enabled when default features are not set.
+    #[cfg(not(feature = "tokio-exporter"))]
+    pub fn build(self) -> Result<PrometheusRecorder, Error> {
```
tobz, line 469: Let's drop the feature gate on this...

metrics-exporter-prometheus/src/lib.rs
```diff
@@ -439,12 +463,31 @@ impl PrometheusBuilder {
...
+    }
+
     /// Builds the recorder and exporter and returns them both.
     ///
     /// In most cases, users should prefer to use [`PrometheusBuilder::install`] to create and
     /// install the recorder and exporter automatically for them.  If a caller is combining
     /// recorders, or needs to schedule the exporter to run in a particular way, this method
     /// provides the flexibility to do so.
+    #[cfg(feature = "tokio-exporter")]
     pub fn build(
```
tobz, line 491: And rename this to `build_with_exporter` while leaving the feature flag.

metrics-exporter-prometheus/src/lib.rs
```diff
@@ -414,6 +437,7 @@ impl PrometheusBuilder {
     ///
     /// An error will be returned if there's an issue with creating the HTTP server or with
     /// installing the recorder as the global recorder.
+    #[cfg(feature = "tokio-exporter")]
     pub fn install(self) -> Result<(), Error> {
         let (recorder, exporter) = self.build()?;
```
tobz, line 442:
```suggestion
        let (recorder, exporter) = self.build_with_exporter()?;

tobz (changes requested):
A few small stylistic nits to match the existing documentation style, and to avoid confusion.

kilpatty:
Okay I think I fixed all those!

tobz (approved):

tobz:
Looks good to me!

I'll merge now and release a new alpha later tonight.

tobz:
Released `metrics-exporter-prometheus@v0.1.0-alpha.7`.

Thanks again for your contribution!

PR #121, opened 2020-11-10
Doc typo

flub:
This fixes a trivial docs typo.  Though also adds a docs build to the CI which will fail if there are warnings while building docs, which happen for e.g. broken links.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,23 @@ jobs:
           override: true
     - name: Run Tests
       run: cargo test
+  docs:
+    runs-on: ubuntu-latest
+    env:
+      RUSTDOCFLAGS: -Dwarnings
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Rust Nightly
+      uses: actions-rs/toolchain@v1
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,23 @@ jobs:
...
+      uses: actions-rs/toolchain@v1
+      with:
+          toolchain: nightly
+          override: true
+          components: rust-docs
+    - name: Check docs
+      uses: actions-rs/cargo@v1
+      with:
+        command: doc
+        args: --workspace --no-deps
```
tobz, line 54: Interesting... do we not have to specify that we want the nightly toolchain for `cargo doc` or is it actually setting it as the default when installing?

tobz: Seems fine regardless, so I'm just gonna approve.

tobz (approved):

.github/workflows/ci.yml
flub: As I understand it the job has no cargo until `actions-rs/toolchain@v1` is run.  After that the toolchain selected with it is the only one available and the default.

Issue #122, opened 2020-11-11
PrometheusBuilder::install panics without a thread-local runtime


str4d:
`PrometheusBuilder::install` calls `PrometheusBuilder::build_with_exporter` to configure the exporter, then creates a `tokio` runtime and uses it to run the exporter.

In https://github.com/metrics-rs/metrics/pull/92 `PrometheusBuilder::build` (now `build_with_exporter`) was refactored to call `Server::try_bind` before starting the server. This means that `hyper` operations happen outside the `exporter` future, whereas previously all `hyper` operations happened inside `Runtime::block_on`.

The problem with this is that `Server::try_bind` eventually calls `tokio::net::tcp::TcpListener::from_std`, which panics if thread-local runtime is not set:
```
    /// # Panics
    ///
    /// This function panics if thread-local runtime is not set.
    ///
    /// The runtime is usually set implicitly when this function is called
    /// from a future driven by a tokio runtime, otherwise runtime can be set
    /// explicitly with [`Handle::enter`](crate::runtime::Handle::enter) function.
```

I believe the solution is to create the runtime first, and then enter its context when building the exporter:
```rust
    pub fn install(self) -> Result<(), Error> {
        let mut runtime = runtime::Builder::new()
            .basic_scheduler()
            .enable_all()
            .build()?;

        let (recorder, exporter) = runtime.enter(|| self.build_with_exporter())?;
        metrics::set_boxed_recorder(Box::new(recorder))?;
```

PR #123, opened 2020-11-11
PrometheusBuilder::install: Build exporter in runtime context

str4d:
Closes https://github.com/metrics-rs/metrics/issues/122


str4d:
Tested this branch with a `[patch.crates-io]` in my local setup, and it resolves my issue.

PR #131, opened 2020-11-18
Add support for culling idle metrics from Prometheus output

### LINKED ISSUE #128: PrometheusExporter: allow dropping idle metrics
Per a discussion in Discord (https://discord.com/channels/500028886025895936/632238397259382827/777930559477252126) it was pointed out that the Prometheus exporter will tend to balloon in memory over time if new metrics (new in the sense of being a yet-unseen name/labels combination) continue to accumulate over time.

We should add the capability to allow a "idle time" to be defined, where metrics that have not been touched within the window will be dropped from storage, and from the registry.

This is two-fold:
- we need to add support to `Registry` to allow dropping the entries (otherwise the next pool would bring them back)
- we need to add the ability to track when a metric was last touched, to know when to drop a metric

Conceptually, we would likely want to store a `key -> (last known value, last update time)` lookup table, where every time we generated a snapshot to respond to a scrape request, we'd check if any of the snapshotted values had changed from the last seen value, and if so, update the time.  If there was no change, and we exceeded our threshold, the metric would be dropped from the registry and the lookup table.

There's probably some details that have to be nailed to make it work cleanly, but overall I don't think it's a particularly complex change to make.

tobz:
Per #128, we wanted the ability to drop/remove idle metrics from output, and from consuming memory, where idleness is defined by how long it has been since the last update to the metric.

This PR takes a stab at adding this support by making a few changes here and there:
- `Registry` now has the concept of generations for a metric, where each operation increments the generation
- `PrometheusRecorder` uses generations to keep a recency list of key -> generation/last update
- when metrics are seen with the same generation as the last time we generated our output, we check their recency
- if the metric has been idle longer than `idle_timeout`, we remove it from the registry, and in turn, the rendered output

Running the basic `metrics-benchmark` test doesn't show any major regressions, and as expected, microbenchmarks for `Registry` have gone up by 6-8% as we're now doing extra work for every operation.  I think this is sustainable, however, as we're still in 60-70ns territory for registry operations.  We expect to hopefully claw some of this time back with future optimizations around key hashing and `dashmap` v4.

Remaining items:
- [x] tests in the exporter itself to exercise dropping metrics

Closes #128


jeromegn:
This is working well for us!

PR #133, opened 2020-11-19
[WIP] deps: update to tokio 0.3

hdevalence:
I made these changes in service of removing old Tokio versions from our dep tree.  Until Hyper 0.14 is released I'm not sure whether you'd want to merge them and accept the git dependency on Hyper.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,3 +9,7 @@ members = [
   "metrics-tracing-context",
 ]
 exclude = ["metrics-observer"]
+
+[patch.crates-io]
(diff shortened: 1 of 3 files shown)


tobz:
Sorry, very delayed response to this.

I would fundamentally be against depending on a specific commit for Hyper, as I dislike Git-based dependencies.  We leave the overall PR open, though, to track upgrading to Hyper 0.14 in general, once they finally publish 0.14.0.

Kononnable:
@hdevalence  Hyper 0.14 was released with tokio 1.0 support. I don't think there were more major changes from 0.3 to 1.0 which would require more code changes here.

strohel:
It looks like tokio has been already updated to 1.0, so this can be closed.

PR #140, opened 2020-12-10
metrics-observer: actually wrote a README

tobz:
As stated in the title, we've added actual documentation for `metrics-observer`, including some screenshots, to demonstrate how and why to use it.


tobz:
Wrong branch.

PR #164, opened 2021-01-25
Tracing context perf smallvec

MOZGIII:
This PR build upon #157.

---

This PR adds another way of filtering labels, and generally rewrites the internal label storage to be more optimal.


tobz:
@MOZGIII Whenever this is ready and you have a chance, it'd be good to have a 2-3 sentence summary of the changes, and some rough before/after numbers.

PR #166, opened 2021-01-30
Prometheus global labels

### LINKED ISSUE #143: [PrometheusBuilder] Global labels for all metrics
I was wondering is it possible to register some global labels to be sent by all of an exporters metrics
eg. for the PrometheusBuilder I'd like to set the application name and version to be sent along with all metrics.

Is this possible today or functionality that may exist I the future?

AshtonKem:
Gives the Prometheus reporter the ability to be configured with global labels, which are then added to each metric reported.

Closes #143.

### DIFF
--- a/metrics-exporter-prometheus/src/builder.rs
+++ b/metrics-exporter-prometheus/src/builder.rs
@@ -431,33 +431,21 @@ mod tests {
         let recorder = PrometheusBuilder::new()
             .add_global_label("foo", "foo")
             .add_global_label("foo", "bar")
-            .add_global_label("bar", "baz")
             .build();
         let key = Key::from(KeyData::from_name("basic_counter"));
         recorder.increment_counter(key, 42);
 
         let handle = recorder.handle();
         let rendered = handle.render();
-        let expected_counter = "# TYPE basic_counter counter\nbasic_counter{foo=\"bar\",bar=\"baz\"} 42\n\n";
+        let expected_counter = "# TYPE basic_counter counter\nbasic_counter{foo=\"bar\"} 42\n\n";
 
         assert_eq!(rendered, expected_counter);
-
-        // let key = Key::from(KeyData::from_name("overridden")
-        //     .with_extra_labels(vec![Label::new("foo", "overridden")]));
-        // recorder.increment_counter(key, 1);
-        //
-        // let rendered = handle.render();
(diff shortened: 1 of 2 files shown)


AshtonKem:
Lesson learned, IntelliJ does not use rustfmt by default.

tobz:
Woo, this is great!  Thank you for starting this PR.

While the structure and logic seem sound, I have some thoughts/concerns about the cloning/allocations that are happening.  I'll review the PR more thoroughly and leave some comments on areas that I think can be polished.

AshtonKem:
Absolutely. I’m new to Rust, so I probably made some noob mistakes. 

I at least think I got the behavior right in terms of what overrides what, and the elimination of duplicates. That’s encoded in the tests, so it should be easy to change under the hood. 

—
Ashton Kemerling

tobz:
Nope, now that I take a closer look, I think the current implementation is actually as good as it can reasonably get.  I wasn't paying enough attention to why certain things are being done the way they are.

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -126,6 +129,24 @@ impl PrometheusBuilder {
...
     }
 
+    /// Adds a label for all metrics recorded.
+    ///
+    /// If called, this provided label will be added to every single metric reported through this
+    /// registry. Repeated calls will add additional labels, duplicate keys will be silently
+    /// overwritten.
+    ///
+    /// If a individual metric reuses a key provided here, the local metric key/value pair will
+    /// win out over the global one for that one call only.
```
tobz, line 139: I'm a bit OCD when it comes to documentation writing:

```suggestion
    /// Adds a global label to this exporter.
    ///
    /// Global labels are applied to all metrics.  Labels defined on the metric key itself have precedence
    /// over any global labels.  If this method is called multiple times, the latest value for a given label
    /// key will be used.

tobz (changes requested):

tobz:
Just one small nit regarding documentation wording, but once that's applied, we should be able to merge.

metrics-exporter-prometheus/src/builder.rs
AshtonKem: Documentation update.

tobz (approved):

tobz:
I'll cut a release later on for this.

tobz:
Wanted to bundle some changes together, but: `metrics-exporter-prometheus@v0.3.0` is now released.

PR #168, opened 2021-02-02
Composite registers

AshtonKem:
Quick PR for "composite" recorders. 

In my day job we recently needed to support two separate metrics locations with Micrometer (InfluxDB & DataDog). This struck me as a reasonable thing that one might need to do either for cost reasons (DD is expensive, so only put key metrics there), or for transition reasons. I'm planning on adding some filtering recorders soon, to make it easy to greenlight specific metrics for specific recorders. 

Docs are really lightweight since I figure you'll want to tweak them to your liking.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
   "metrics",
   "metrics-macros",
   "metrics-util",
+  "metrics-exporter-composite",
   "metrics-exporter-tcp",
   "metrics-exporter-prometheus",
   "metrics-tracing-context",
--- a/README.md
+++ b/README.md
@@ -49,24 +49,36 @@ Overall, this repository is home to the following crates:
 
 * [`metrics`][metrics]: A lightweight metrics facade, similar to [`log`][log].
 * [`metrics-macros`][metrics-macros]: Procedural macros that power `metrics`.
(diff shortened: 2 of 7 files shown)


AshtonKem:
Really not sure why the docs are unhappy

tobz:
You're gonna hate me, but: this already exists. :)

https://docs.rs/metrics-util/0.5.0/metrics_util/layers/index.html

The layers machinery includes a trait that lets us compose layers on top of each other, as well as a helper type `Stack` for acting as the root layer which can be installed as the global recorder, and some pre-existing layers that make sense, two of which are `Fanout`, which does what your composite exporter is doing, and `Filter`, which allows filtering metrics by name.

The filtering layer might not be exactly what you want need, but `Fanout` should be, and you can simply write any other layer that's required.

If you want to transform this PR to be building some other generic layer that you need, that's fine, but otherwise I'll close it in a day or two if you haven't already.

AshtonKem:
Ha! No worries. This seems like something that can be improved by a bit of better documentation so that everyone else can find the `Fanout` functionality. 

Either way, I learned a lot writing this PR, so not a loss.

tobz:
👍🏻 

We definitely still have a big docs push to make before landing at 1.0, so I'll make a note that the layer stuff isn't well documented.

Issue #165, opened 2021-01-30
The macros no longer support dynamic keys


Marwes:
Why was this removed in 0.13? It is still possible to call the recorder methods manually but constructing the Key is rather annoying. Used dynamic keys for things like:
```rust
metrics::counter!(format!("response_status_{}", response.status().as_u16()), 1);
```

tobz:
Well, the simple answer is: practically speaking, we never designed it to be specifically allowed, and it working as such was mostly a side effect of all of the `Into` impls we had.  The example you gave would normally be achieved with something like:

```rust
metrics::increment!("http_responses", "status" => response.status().as_u16().as_str());
```

There's probably a way to bring this back: `Key`/`KeyData` definitely support initialization from owned strings.  The only current hurdle is the macros themselves, and the arguments they expect.  I'll have to think about it.

tobz:
I've been toying with the macro stuff, and this is totally doable.  I'm still in the process of going through and adding a lot more unit tests to make sure this works as intended in all cases.

You can follow the PR: #170

tobz:
Merged and released.  I cut all new minors for essentially all of the crates.  Given the new tests that were added, there's fairly solid coverage of the macros, but let me know if anything seems amiss.

PR #170, opened 2021-02-02
Allow passing owned strings as metric key.

tobz:
Previously, due to the way the macros were written, users could specify a metric key as either a string literal or an expression that pointed to an owned string, due to the `Into<T>` implementations for `Key` and the way the macros handled the input arguments.  While not an explicit design goal, this was broken in the rewrite for 0.13: #165

This PR reworks the macros such that they can properly handle either a string literal or owned string as the metric key.  Minimal work was required on the `Key`/`KeyData` side as we mostly had a sufficient set of `from_*` methods.  The real work is in the macro argument parsing and code generation itself, dealing with the matrix of combinations: is the name static, are the labels static, and all the combinations in thereof.


tobz:
CI breakage looks to be related to Alexhuszagh/rust-lexical#55 so I'm trying to get that sorted.

PR #177, opened 2021-02-12
impl Ord/PartialOrd for key types

jmagnuson:
Allows for interoperability with structures like [`LruCache`](https://docs.rs/lru_time_cache/0.11.7/lru_time_cache/struct.LruCache.html).

PR #178, opened 2021-02-16
add ordering unit test for Key

jmagnuson:
A follow-up to #177; `Ord` and `PartialOrd` were manually implemented for `Key`, so unit tests for them are probably warranted.


tobz:
Nothing wrong with a good old-fashioned unit test. Danke!

tobz (approved):

PR #184, opened 2021-03-15
Relax atomic operations of `update_gauge` and `increment_counter`

blt:
This commit relaxes the atomics of the aforementioned functions, removing the sync pressure on machines with large numbers of CPUs or ARM. As the reads of counters and gauges were formerly relaxed I believe this maintains the semantic intentions, though it is now possible for a writer to immediately read and "lose" the write temporarily if the read is done from a different CPU, in the case of the counter.

I noticed this while digging through https://github.com/timberio/vector/issues/6770. I haven't managed to test in our project because of https://github.com/timberio/vector/issues/6412.

Signed-off-by: Brian L. Troutwine <brian@troutwine.us>

### DIFF
--- a/metrics-util/src/handle.rs
+++ b/metrics-util/src/handle.rs
@@ -48,7 +48,7 @@ impl Handle {
     pub fn increment_counter(&self, value: u64) {
         match self {
             Handle::Counter(counter) => {
-                counter.fetch_add(value, Ordering::SeqCst);
+                counter.fetch_add(value, Ordering::Relaxed);
             }
             _ => panic!("tried to increment as counter"),
         }
@@ -60,7 +60,7 @@ impl Handle {
     pub fn update_gauge(&self, value: GaugeValue) {
         match self {
             Handle::Gauge(gauge) => {
-                let _ = gauge.fetch_update(Ordering::SeqCst, Ordering::SeqCst, |curr| {
+                let _ = gauge.fetch_update(Ordering::AcqRel, Ordering::Relaxed, |curr| {
                     let input = f64::from_bits(curr);
                     let output = value.update_value(input);
                     Some(output.to_bits())


metrics-util/src/handle.rs
```diff
@@ -60,7 +60,7 @@ impl Handle {
     pub fn update_gauge(&self, value: GaugeValue) {
         match self {
             Handle::Gauge(gauge) => {
-                let _ = gauge.fetch_update(Ordering::SeqCst, Ordering::SeqCst, |curr| {
+                let _ = gauge.fetch_update(Ordering::AcqRel, Ordering::Relaxed, |curr| {
```
tobz, line 63: This needs to be `Ordering::Acquire` on the failure ordering.

If another update happened that caused this line to hit the failure case, it would be loaded with `Relaxed`, which from my understanding of the orderings, would potentially allow for a stale read.

tobz (changes requested):
This all seems reasonable; just a single spot we need to tweak.

metrics-util/src/handle.rs
blt: I don't think that interpretation is correct, but I admit I found the documentation for `fetch_update` confusing. The implementation of `fetch_update` from [here](https://doc.rust-lang.org/src/core/sync/atomic.rs.html#867) is

```rust
pub fn fetch_update<F>(
    &self,
    set_order: Ordering,
    fetch_order: Ordering,
    mut f: F,
) -> Result<bool, bool>
where
    F: FnMut(bool) -> Option<bool>,
{
    let mut prev = self.load(fetch_order);
    while let Some(next) = f(prev) {
        match self.compare_exchange_weak(prev, next, set_order, fetch_order) {
            x @ Ok(_) => return x,
            Err(next_prev) => prev = next_prev,
        }
   }
   Err(prev)
}
```

The load is indeed done with `fetch_order` ordering but the underlying `compare_exchange_weak` will mean that if/when a stale read is made that read is never the foundation of a write. This would not be true if `set_order` were also relaxed. If a thread in `fetch_update`'s loop reads an out of date value it will loop around again. What with `_weak` already being subject to spurious failures we'll be hard-pressed to distinguish those additional loops from what a more strict ordering might allow, the downside of more strict ordering being we're saving cheap loops to avoid expensive cross-CPU synchronization. 

So, I guess it depends on what you mean by "stale". If the concern is losing potential writes with relaxed ordering that won't happen, though it is possible calls will loop a little more while avoiding expensive cross-CPU syncs.

tobz: Yeah, you're right, I was fixated on the "`prev` gets set to the value of a relaxed load, oh no!" but ignoring that it's still using that value for the CMPXCHG comparison which would resolve any discrepancies, as you've pointed out. 👍🏻

tobz (approved):

tobz:
Merged. Thanks for you contribution!

Issue #187, opened 2021-04-01
prometheus_server example is broken


sandrovicente:
If we run `prometheus_server` example (` RUST_BACKTRACE=1 cargo run -p metrics-exporter-prometheus --example prometheus_server`), we get:

```
thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', /home/sandrov/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.4.0/src/runtime/context.rs:18:26
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07e0e2ec268c140e607e1ac7f49f145612d0f597/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/07e0e2ec268c140e607e1ac7f49f145612d0f597/library/core/src/panicking.rs:92:14
   2: core::option::expect_failed
             at /rustc/07e0e2ec268c140e607e1ac7f49f145612d0f597/library/core/src/option.rs:1321:5
   3: core::option::Option<T>::expect
             at /home/sandrov/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:349:21
   4: tokio::runtime::context::io_handle::{{closure}}
             at /home/sandrov/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.4.0/src/runtime/context.rs:18:13
   5: std::thread::local::LocalKey<T>::try_with
             at /home/sandrov/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:272:16
   6: std::thread::local::LocalKey<T>::with
             at /home/sandrov/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:248:9
   7: tokio::runtime::context::io_handle
             at /home/sandrov/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.4.0/src/runtime/context.rs:16:9
   8: tokio::io::driver::Handle::current
             at /home/sandrov/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.4.0/src/io/driver/mod.rs:262:13
   9: tokio::io::poll_evented::PollEvented<E>::new_with_interest
[...]

PR #189, opened 2021-04-07
Fix broken prometheus example.

BratSinot:
Related with #187 .

### DIFF
--- a/metrics-exporter-prometheus/examples/prometheus_server.rs
+++ b/metrics-exporter-prometheus/examples/prometheus_server.rs
@@ -11,8 +11,7 @@ use metrics_util::MetricKindMask;
 use quanta::Clock;
 use rand::{thread_rng, Rng};
 
-#[tokio::main]
-async fn main() {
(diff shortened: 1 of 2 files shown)


tobz:
Per my [comment](https://github.com/metrics-rs/metrics/issues/187#issuecomment-814954795) on the issue, this should actually be fixing the runtime enter guard in `PrometheusBuilder::install`.

BratSinot:
> Per my [comment](https://github.com/metrics-rs/metrics/issues/187#issuecomment-814954795) on the issue, this should actually be fixing the runtime enter guard in `PrometheusBuilder::install`.

Oh, yeah. I still can't quite get used to Rust async's.
Saved guard, as @sandrovicente suggests.

tobz:
And just to make sure: does the `prometheus_server` example actually run cleanly now?

tobz:
I checked this, it works.

tobz:
@BratSinot Thanks for your contribution!

PR #194, opened 2021-04-11
Experiment: switch to `&Key` for `Recorder`

tobz:
In timberio/vector#7014, @blt [remarked](https://github.com/timberio/vector/pull/7014#discussion_r608767945) that they spend a lot of time creating temporary `Key` instances, to the point it shows up in profiling.  This gave me an idea that we could, in the static case, avoid this entirely by passing `&Key` around instead of `Key`.

In this PR, we're doing two major things:
- `Recorder` now takes `&Key` instead of `Key`
- `Registry` has been reworked/optimized to support `&Key` in the process

Obviously, we've reworked a lot of boundaries/`Recorder` implementations (i.e. layers) to support the switch to `&Key`.  In turn, `KeyData` is now `Key`.  As now we only have `&Key` to work with, we had to modify `Registry` as well.  Rather than clone the key every time and continue using the `dashmap`/entry API approach, we've reworked `Registry`.

Given the primary use case, we've encoded the metric kind itself into `Registry`, so operating on the registry requires specifying a reference to the key type _and_ the metric kind.  This lets us avoid having to rewrap the key in a new type (another allocation!) but we also avoid having to hash the kind, instead transforming it to an index we use for a fast lookup.

Long story short, we've made `Registry` significantly faster -- 45 to 55% faster --  and we've reduced the number of temporaries required for the optimized static path.

### DIFF
--- a/metrics-util/src/registry.rs
+++ b/metrics-util/src/registry.rs
@@ -2,8 +2,16 @@ use core::{
     hash::Hash,
     sync::atomic::{AtomicUsize, Ordering},
 };
-use dashmap::DashMap;
-use std::collections::HashMap;
+use std::{hash::{BuildHasherDefault, Hasher}, iter::repeat};
+
+use hashbrown::{HashMap, hash_map::RawEntryMut};
+use parking_lot::RwLock;
+use t1ha::T1haHasher;
+
+use crate::MetricKind;
+
+type RegistryHasher = T1haHasher;
+type RegistryHashMap<K, V> = HashMap<K, Generational<V>, BuildHasherDefault<RegistryHasher>>;
 
 /// Generation counter.
 ///
@@ -58,7 +66,8 @@ where
     K: Eq + Hash + Clone + 'static,
     H: 'static,
 {
-    map: DashMap<K, Generational<H>>,
+    shards: Vec<Vec<RwLock<RegistryHashMap<K, H>>>>,
+    mask: usize,
 }
 
 impl<K, H> Registry<K, H>
@@ -68,9 +77,44 @@ where
 {
     /// Creates a new `Registry`.
     pub fn new() -> Self {
+        let shard_count = std::cmp::max(1, num_cpus::get()).next_power_of_two();
+        let mask = shard_count - 1;
+        let counters = repeat(())
+            .take(shard_count)
+            .map(|_| RwLock::new(RegistryHashMap::default()))
+            .collect();
+        let gauges = repeat(())
+            .take(shard_count)
+            .map(|_| RwLock::new(RegistryHashMap::default()))
+            .collect();
+        let histograms = repeat(())
+            .take(shard_count)
+            .map(|_| RwLock::new(RegistryHashMap::default()))
+            .collect();
+
+        let shards = vec![counters, gauges, histograms];
+
         Self {
-            map: DashMap::new(),
+            shards,
+            mask,
         }
+    }
+
+    #[inline]
+    fn get_hash_and_shard(&self, kind: MetricKind, key: &K) -> (u64, &RwLock<RegistryHashMap<K, H>>) {
+        let mut hasher = RegistryHasher::default();
+        let hash = hash_key(&mut hasher, key);
+
+        // SAFETY: We map each MetricKind variant -- three at present -- to a usize value
+        // representing an index in a vector, so we statically know that we're always extracting our
+        // sub-shards correctly.  Secondly, we initialize vector of subshards with a power-of-two
+        // value, and `self.mask` is `self.shards.len() - 1`, thus we can never have a result from
+        // the masking operation that results in a value which is not in bounds of our subshards
+        // vector.
+        let shards = unsafe { self.shards.get_unchecked(kind_to_idx(kind)) };
+        let shard = unsafe { shards.get_unchecked(hash as usize & self.mask) };
+
+        (hash, shard)
     }
 
     /// Perform an operation on a given key.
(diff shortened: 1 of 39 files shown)


metrics-util/src/registry.rs
```diff
@@ -2,8 +2,16 @@ use core::{
     hash::Hash,
     sync::atomic::{AtomicUsize, Ordering},
 };
-use dashmap::DashMap;
-use std::collections::HashMap;
+use std::{hash::{BuildHasherDefault, Hasher}, iter::repeat};
+
+use hashbrown::{HashMap, hash_map::RawEntryMut};
+use parking_lot::RwLock;
+use t1ha::T1haHasher;
```
blt, line 9: TIL about t1ha. Neat.

blt:
This looks extremely promising. I'll try it out in vector once this is ready but our custom fork of registry -- we removed the generation logic -- will muddy the comparison somewhat. The fork we did is a little before my time but my understanding is we cut that to avoid paying the cost of keeping those counters updated; we never delete any keys. More an FYI than anything. I would be interested in noodling over how to support a registry that's append-only in this project since surely we're not the only people that never delete keys but that's outside the scope of this PR for sure. 

I'm very enthusiastic about this PR.

tobz:
@blt Open to suggestions on changing `Registry` to support avoid using generations.  Maybe like multiple impl blocks? Off the top of my head, I forget where specialization support in self-impl blocks stands.  I'm sure we can work _something_ up that supports the `metrics` use case and the `vector` use case.

blt:
> @blt Open to suggestions on changing `Registry` to support not using generations. Maybe like multiple impl blocks? Off the top of my head, I forget where specialization support in self-impl blocks stands.

I want to say that's still nightly only? I will check. 

> I'm sure we can work _something_ up that supports the `metrics` use case and the `vector` use case.

I think if `Generational` were a trait that'd work for us. Then we could implement a no-op variant. It'd be less clean than a specialization approach -- deletion would still exist in the API but never do anything -- but it would work for our use case.

blt:
> I want to say that's still nightly only? I will check.

Hmm, well, so far as I can tell not much has changed recently. `min_specialization` is now available for use in stdlib but that's about it. I'd be very glad to learn I'd missed something though.

tobz:
So yeah, maybe we can look into a trait-based approach.  That'd work for me.

tobz:
Things are looking _very_ fast at this point:
flamegraph

We're basically limited by the throughput of having to look up the metric in the registry itself.  We likely can't do any better beyond handing out dedicated handles -- i.e. `Arc<Handle>` -- that could be held locally, but we also can't do that with the macro scheme anyways, so I would say we're about as efficient as we possibly can be here.

Practically speaking, the `metrics-benchmark` prior to this benchmark was running at roughly ~14M ops/sec.  After the PR, we're at ~34M ops/sec.  A big chunk of that was that holding the raw histogram values in memory was slowwww, so we've written a custom benchmark recorder implementation where we never actually pull out all the values from `AtomicBucket<T>`, but simply periodically clear it to simulate a normal exporter configuration.  That change got us from ~14M ops/sec to around ~24M ops/sec.

The biggest limitation with this set-up is that we've bumped up the size of `Key` (neé `KeyData`) by two words, for an atomic boolean and `u64`, respectively, where we memoize the hash state.  We can't const hash in our const construction of static keys, due to not being able to take our CoW pointer back to a `&str`/`&[T]` in a `const` fashion, at least on stable.  Once that was available on stable, we could actually switch to hashing entirely at construction, which would shrink it back down to a single word, and avoid the atomics entirely.  It's not exactly _slow_ now, it's just still slightly suboptimal.

blt:
> We're basically limited by the throughput of having to look up the metric in the registry itself. We likely can't do any better beyond handing out dedicated handles -- i.e. `Arc<Handle>` -- that could be held locally, but we also can't do that with the macro scheme anyways, so I would say we're about as efficient as we possibly can be here.

This strikes me as an acceptable situation. I'm going to push vector to use an `Arc<Handle>` as much as possible and I think there's only a few places where that won't work. 
 
> Practically speaking, the `metrics-benchmark` prior to this benchmark was running at roughly ~14M ops/sec. After the PR, we're at ~34M ops/sec. A big chunk of that was that holding the raw histogram values in memory was slowwww, so we've written a custom benchmark recorder implementation where we never actually pull out all the values from `AtomicBucket<T>`, but simply periodically clear it to simulate a normal exporter configuration. That change got us from ~14M ops/sec to around ~24M ops/sec.

Nice. 

> The biggest limitation with this set-up is that we've bumped up the size of `Key` (neé `KeyData`) by two words, for an atomic boolean and `u64`, respectively, where we memoize the hash state. We can't const hash in our const construction of static keys, due to not being able to take our CoW pointer back to a `&str`/`&[T]` in a `const` fashion, at least on stable. Once that was available on stable, we could actually switch to hashing entirely at construction, which would shrink it back down to a single word, and avoid the atomics entirely. It's not exactly _slow_ now, it's just still slightly suboptimal.

This also strikes me as acceptable, as a user of the library. I'd follow a ticket on this project that tracks whatever rustc stabilization ticket exists. While I _would_ like a smaller `Key` I'm not blocked on that need. IIRC vector has a few hundred metrics in flight tops.

tobz:
The one thing I haven't done is run Massif/DHAT/etc to look at the memory usage, but intuitively I would expect slightly better performance given the lack of `Key` temporaries.

blt:
> The one thing I haven't done is run Massif/DHAT/etc to look at the memory usage, but intuitively I would expect slightly better performance given the lack of `Key` temporaries.

If you have a program to run that's common across master branch versus this PR I'm happy to run it under massif.

tobz:
You _could_ run `metrics-benchmark` but given the bit about the new benchmark-specific recorder, I'd imagine the profiles might be impractically different for the purpose of distinguishing behavior.  Then again, I'm not familiar enough with them: is the tracking granular enough to see allocations by specific type?  If so, then it should be representative.

blt:
> You _could_ run `metrics-benchmark` but given the bit about the new benchmark-specific recorder, I'd imagine the profiles might be impractically different for the purpose of distinguishing behavior.

Hmm, alright. A benchmark might end up looking kind of goofy under analysis since we'd pick up whatever criterion is doing, but I can take that and make hay with it. 

> Then again, I'm not familiar enough with them: is the tracking granular enough to see allocations by specific type? If so, then it should be representative.

Hmm, maybe? Massif intercepts any malloc and keeps count of what called malloc and how often. DHAT does the same but keeps track of how often those allocations are accessed. The big question is how well massif/DHAT end up assigning allocations to something we recognize in the program. 

I'll put something together right quick and we can have a look.

tobz:
FWIW, `metrics-benchmark` is really just a producer/consumer thread benchmark, not a microbenchmark.  No Criterion at all, or any other similar crate.

blt:
> FWIW, `metrics-benchmark` is really just a producer/consumer thread benchmark, not a microbenchmark. No Criterion at all, or any other similar crate.

D'oh! So it is. Got fooled by the name. :)

blt:
I'm comparing 7935f2faf66844596a708d040952eebb851a020d `metric-benchmarks` with current main, command:

`> valgrind --tool=massif ./target/release/metrics-benchmark --duration 60 --producers 32`

I'll post results here, with screenshots, as I have them.

blt:
Huh, well. I think `metrics-benchmark` running under massif has an odd interaction with the histogram implementation. See screenshots:

keyref-branch-massif
main-branch-massif

[massif-main-branch.out.zip](https://github.com/metrics-rs/metrics/files/6321604/massif-main-branch.out.zip)
[massif-keyref-branch.out.zip](https://github.com/metrics-rs/metrics/files/6321608/massif-keyref-branch.out.zip)

Run under massif these programs never terminate either; I killed them manually after 5 minutes. So, something weird's going on and I'm not sure how useful these results are tbh.

tobz:
Fair, I think the PR is probably worth merging for the performance alone.  I can take a look later with a Vector benchmark from the instructions you've sprinkled in various PRs. 👍🏻

blt:
Similar story for DHAT:

Screenshot_2021-04-15 DHAT Viewer - dhat out 3437501 - Total (blocks)
Screenshot_2021-04-15 DHAT Viewer - dhat out 3437543 - Total (blocks)

Honestly not sure what to make of this! Kind of seems like an easier way to trigger the bug that got side-stepped in https://github.com/timberio/vector/pull/7014.

blt:
> Fair, I think the PR is probably worth merging for the performance alone. I can take a look later with a Vector benchmark from the instructions you've sprinkled in various PRs. 👍🏻

Fully onboard. There's some... peculiar interaction with valgrind and metrics-rs' histogram but I don't see that as a blocker here.

PR #195, opened 2021-04-14
Fix name of constant in macro expansion

twz123:
One macro branch errorneously referenced METRIC_LABELS as METRICS_LABELS, leading to compilation failures.


tobz:
Ah, good catch!  Looks like we need a high-level test for dynamic key + static labels.

tobz (approved):

tobz:

PR #198, opened 2021-04-14
Support non-literal names along with literal labels in macros

### LINKED ISSUE #196: Add macro test for dynamic keys + static labels.
As shown in #195, we had a bug where a dynamic name/static labels combination had compilation errors due to incorrect functionality in the procedural macros.  We need to add a test to stress this use case so that we don't miss it in tests in the future.

twz123:
Arrays of Labels don't implement `IntoLabels`. They have already a static lifetime and can be borrowed instead of owned. Introduce `KeyedData::from_static_labels(..)` to support the use case of a non-literal name and a literal label set. Also add tests for that case.

Fixes #196.

Without the change, the newly added test cases fail with

```
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error[E0277]: the trait bound `Label: From<&Label>` is not satisfied
   --> $DIR/01_basic.rs:18:5
    |
18  |     counter!(format!("response_status_{}", some_u16), 1, "uvw" => "xyz");
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&Label>` is not implemented for `Label`
    |
   ::: $WORKSPACE/metrics/src/key.rs
    |
    |         L: IntoLabels,
    |            ---------- required by this bound in `KeyData::from_parts`
    |
    = help: the following implementations were found:
              <Label as From<&(K, V)>>
    = note: required because of the requirements on the impl of `Into<Label>` for `&Label`
    = note: required because of the requirements on the impl of `IntoLabels` for `&[Label; 1]`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
```

### DIFF
--- a/metrics-macros/src/lib.rs
+++ b/metrics-macros/src/lib.rs
@@ -425,7 +425,7 @@ fn generate_metric_key(name: &Expr, labels: &Option<Labels>) -> TokenStream2 {
         // to figure out the correct key.
         if has_labels {
             quote! {
-                metrics::Key::Owned(metrics::KeyData::from_parts(#name, & METRIC_LABELS))
+                metrics::Key::Owned(metrics::KeyData::from_static_labels(#name, &METRIC_LABELS))
             }
         } else {
             quote! {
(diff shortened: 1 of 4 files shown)


tobz:
Sweet! This looks good to me. 👍🏻

PR #204, opened 2021-05-11
metrics-exporter-prometheus: Add support for subnet allow list

str4d:
This enables access to the Prometheus scrape endpoint to be restricted to a defined set of IPs and/or subnets.

Note that this on its own is insufficient for access control, if the exporter is running in an environment alongside applications (such as web browsers) that are susceptible to [DNS rebinding attacks](https://en.wikipedia.org/wiki/DNS_rebinding). I plan to add support for HTTP auth in a subsequent PR.

### DIFF
--- a/metrics-exporter-prometheus/CHANGELOG.md
+++ b/metrics-exporter-prometheus/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- next-header -->
 
 ## [Unreleased] - ReleaseDate
+### Added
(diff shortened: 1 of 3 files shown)


tobz:
Sorry for the delay in reviewing this, been on vacation between jobs. 😎 🌴 

Everything looks good to me for this particular PR.  If you're looking to add more functionality, I'd say that you should open an issue first so we can discuss it.  As soon as we start talking about HTTP authentication, then we're potentially talking about TLS, and so I want to make sure we aren't slowly turning this particular exporter into a kitchen sink of all possible functionality.

PR #209, opened 2021-05-24
Fix minium version of crossbeam-epoch

flub:
The Atomic::compare_exchange function only appears in 0.9.2.


tobz:
Ah yes, woops!  Thanks for fixing this. :)

tobz (approved):

Issue #215, opened 2021-06-21
support prometheus push gateway


NOBLES5E:
> Prometheus client libraries should have a feature to push the registered metrics to a Pushgateway.

https://github.com/prometheus/pushgateway

NOBLES5E:
@tobz I'd like to help with the implementation if desired. But I'm not sure whether you think it should be implementation here.

tobz:
@NOBLES5E I can't explicitly think of a reason why we _wouldn't_ want push gateway support in `metrics-exporter-prometheus`.  Feel free to start working on that, and you can reach out back here or in our [Discord](https://discord.gg/eTwKyY9) if you have questions or need help.

NOBLES5E:
@tobz Thanks! Just created a draft PR here: https://github.com/metrics-rs/metrics/pull/217.

PR #217, opened 2021-07-06
feat: add support for prometheus push gateway

NOBLES5E:
close https://github.com/metrics-rs/metrics/issues/215

### DIFF
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -18,7 +18,7 @@ keywords = ["metrics", "telemetry", "prometheus"]
 [features]
 default = ["tokio-exporter"]
 tokio-exporter = ["hyper", "ipnet", "tokio"]
-push-gateway = ["reqwest"]
+push-gateway = ["reqwest", "tracing"]
 
 [dependencies]
 metrics = { version = "^0.16", path = "../metrics" }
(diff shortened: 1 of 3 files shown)


tobz:
Overall, the approach here seems pretty straightforward.

I'm currently coming back off vacation and will need a few days to review this properly.

tobz (approved):
Overall, this looks good to me.

It would be nice if there was a way to use `hyper` for the push gateway as well to reduce the dependency count.  Out of curiosity, did you try `hyper` before using `reqwest`?  The only thing I can think of that `reqwest` might be handling automatically is something like TLS, but I wonder how hard that'd be to use with pure `hyper`.

I'm happy to merge it as-is, but just curious to hear your thoughts.

NOBLES5E:
@tobz Thanks for your feedback. I tried to use hyper client instead of reqwest, but later found that to deal with https in hyper we need to depend on another crate `hyper_tls`, and

 1. parse the url to see if it is "http" or "https"
 2. parse http response to see if it redirects to https
 3. do the final request
 
I am not confident enough that I have considered all the cases, so I just use reqwest instead..

NOBLES5E:
@tobz I just added an example program using prometheus push gateway for completeness.

tobz (approved):
Yep, makes sense re: `hyper` vs `reqwest`.  Just wanted to make sure we weren't ignoring a potential smaller change.  Thanks for giving it a try. 👍🏻 

Everything else looks good to me.  I'll merge once the CI passes.

tobz:
@NOBLES5E Thanks again for your contribution! 👍🏻

tobz:
Released as `metrics-exporter-prometheus@v0.6.0`.

PR #284, opened 2022-03-05
allow for `Storage` with internal state

nrxus:
This will let users of the `Registry` create custom structs that
implement `Storage` where the instances for the metrics are dependent
on data it knows about the world and the key for that metric

In particular my use case for this will be that sometimes I want to create `Summary` and sometimes I want to create `Histogram` when a `metrics::histogram!` metric is emitted. Currently the storages out of the box are insufficient because the "histogram" storage stores a boundless `Vec<f64>` whereas it would be more efficient in a "true histogram" case to just keep track of the counts for each bucket and then merge the histograms in each shard when collecting metrics.


tobz (changes requested):
Overall, seems reasonable.  Still need to take a look at the benchmarks but left some comments in the meantime.

tobz:
No worries at all. 👍🏻 I'll take another look today.

tobz:
This is released in `metrics-util@0v0.12.0`.

PR #285, opened 2022-03-08
Add various exports.

TimonPost:
Adds various missing exports.


tobz:
Sorry, I realized there was a commit that I had meant to push this morning, and it did some of the stuff you have in your PR, but also moved some exports around.  Just pushed it now so that's the cause of the conflict here.

TimonPost:
Cool then I close this one, wasn't to much work. Have you also exposed the: distribution, distribution builder and the sanitize functions? I have not seen those yet. Would like to be able to use those in my project. An other question, when are you planning to release those changes?

tobz:
It doesn't expose anything related to the Prometheus exporter, so feel free to reopen this: I have a few thoughts on how I'd want the exports structured.

I'm happy to cut a new release of `metrics-util` whenever but if you also want some new exports in `metrics-exporter-prometheus`, then I'd rather get this/whatever PR merged in first so I can do the releases all at once.

TimonPost:
Just updated the PR.

Would be really nice if there is a new release within this week. As im working on some project and hope to open a PR for it this week. Otherwise I will have to copy logic which feels a bit bad.

metrics-exporter-prometheus/src/lib.rs
```diff
@@ -101,9 +101,10 @@
 #![deny(missing_docs)]
 #![cfg_attr(docsrs, feature(doc_cfg), deny(rustdoc::broken_intra_doc_links))]
 mod common;
-pub use self::common::{BuildError, Matcher};
+pub use self::common::{BuildError, Matcher, sanitize_description, sanitize_label_key, sanitize_label_value, sanitize_metric_name};
```
tobz, line 104: I think what I'd want to do here is move all of these `sanitize_*` free functions into a new module called `rendering` at the top level, and then expose the module publicly.

That gives us a place to expose more of the rendering code in the future without cluttering the top level of the crate, in terms of what the docs end up looking like.

tobz (changes requested):

metrics-exporter-prometheus/src/lib.rs
TimonPost: Good idea. I can do that.

TimonPost: Maybe we can call it formatting @tobz ?

tobz: Yeah, I'm fine with `formatting`.

TimonPost: I also moved the write_line functions to this module. They all have the same thing in common, formatting strings and sanitizing them for Prometheus.

tobz:
Looks like this needs `cargo fmt` run on it, and the test failures need to be addressed.

TimonPost:
Pushed to fix some of the compile errors. Locally it works now, CI is waiting for your approval.

TimonPost:
For now, I did not expose:

```
invalid_label_key_character,
invalid_label_key_start_character,
invalid_metric_name_character, 
invalid_metric_name_start_character, 
sanitize_label_value_or_descpiption
```

Since they did not seem very relevant and were mostly used by the relevant functions:

```
sanitize_description
sanitize_label_value
sanitize_label_key
sanitize_metric_name
write_metric_line
write_type_line
write_help_line
key_to_parts
```

tobz:
This is released in `metrics-exporter-prometheus@0v0.9.0`.

TimonPost:
Awesome @tobz !

PR #288, opened 2022-03-16
metrics-tracing-context: add Allowlist label filter

look:
In many metrics system (for example, Datadog) sending all tags will be cost prohibitive if not impossible. Since filtering metrics tags to a small set of allowed tags is a common requirement, add a `LabelFilter` to support this easily.

**Note:** I'm fairly new with Rust, please let me know if there are better ways to implement this.

### DIFF
--- a/metrics-tracing-context/src/label_filter.rs
+++ b/metrics-tracing-context/src/label_filter.rs
@@ -1,5 +1,7 @@
 //! Label filtering.
 
+use std::collections::HashSet;
+
 use metrics::Label;
 
 /// [`LabelFilter`] trait encapsulates the ability to filter labels, i.e.
@@ -16,5 +18,26 @@ pub struct IncludeAll;
 impl LabelFilter for IncludeAll {
     fn should_include_label(&self, _label: &Label) -> bool {
         true
+    }
+}
+
+/// A [`LabelFilter`] that allows an allowed list of label names.
+#[derive(Debug, Clone)]
+pub struct Allowlist {
+    /// The set of allowed label names.
+    label_names: HashSet<String>,
+}
+
+impl Allowlist {
+    /// Create a [`Allowlist`] filter with the provide label names.
+    pub fn new(label_names: &[&str]) -> Allowlist {
+        let set: HashSet<String> = label_names.iter().map(|l| l.to_string()).collect();
+        Allowlist { label_names: set }
+    }
+}
+
+impl LabelFilter for Allowlist {
+    fn should_include_label(&self, label: &Label) -> bool {
+        self.label_names.contains(label.key())
     }
 }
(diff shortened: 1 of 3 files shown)


metrics-tracing-context/src/label_filter.rs
```diff
@@ -18,3 +20,24 @@ impl LabelFilter for IncludeAll {
...
+
+/// A [`LabelFilter`] that allows an allowed list of label names.
+#[derive(Debug, Clone)]
+pub struct Allowlist {
+    /// The set of allowed label names.
+    label_names: HashSet<String>,
+}
+
+impl Allowlist {
+    /// Create a [`Allowlist`] filter with the provide label names.
```
tobz, line 32:
```suggestion
    /// Create a [`Allowlist`] filter with the provided label names.
```

metrics-tracing-context/src/lib.rs
```diff
@@ -125,6 +125,14 @@ impl TracingContextLayer<label_filter::IncludeAll> {
     }
 }
 
+impl TracingContextLayer<label_filter::Allowlist> {
+    /// Creates a new [`TracingContextLayer`] with an allowed list of label
+    /// names.
+    pub fn allowlist(label_names: &[&str]) -> Self {
```
tobz, line 131: A naming and doc comment tweak to make it match the suggestions above:

```suggestion
    /// Creates a new [`TracingContextLayer`] that only allows labels contained
    /// in a predefined list.
    pub fn only_allow<I, S>(allowed: I) -> Self
    where
        I: IntoIterator<Item = S>,
        S: AsRef<str>,
    {
```

tobz (changes requested):
Overall, seems like a reasonable addition!

Left some suggestions just to tighten up the language and API.

metrics-tracing-context/src/label_filter.rs
```diff
@@ -18,3 +20,24 @@ impl LabelFilter for IncludeAll {
...
+    pub fn new(label_names: &[&str]) -> Allowlist {
```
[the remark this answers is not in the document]
look: Thanks, I was trying to get `IntoIterator` working but was running into some problems!

metrics-tracing-context/tests/integration.rs
```diff
@@ -344,3 +348,34 @@ fn test_label_filtering() {
         )]
     )
 }
+
+#[test]
+fn test_label_allowlist() {
+    let (_guard, snapshotter) = setup(TracingContextLayer::allowlist(&["env", "service"]));
```
look, line 354:
```suggestion
    let (_guard, snapshotter) = setup(TracingContextLayer::only_allow(&["env", "service"]));
```

look (commented):
Thanks for the feedback!

tobz (approved):

tobz:

tobz:
Released in `metrics-tracing-context@v0.11.0`.

PR #290, opened 2022-03-18
metrics-exporter-prometheus: properly sanitize metric names

### LINKED ISSUE #289: prometheus exporter sanitizes valid metric names
I have a metric `compaction_level2_duration_ns` that is sanitized into `compaction_level__duration_ns`.  I believe this metric is a valid [prometheus name.](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels).

Here is a test that shows that it fails:

```rust
    #[test]
    fn check_sanitize() {
        let name = "compaction_level2_duration_ns".to_owned();
        let actual = sanitize_metric_name(&name);
        assert_eq!(name, actual);
    }
```

The sanitize code is here:

https://github.com/metrics-rs/metrics/blob/5f720a0a8f53aaee4bfa9de1d9fc470f1ed6b3c3/metrics-exporter-prometheus/src/formatting.rs#L111-L115

I think it is because it replaces the first n occurrences matching the predicate, regardless of their position.

gnuvince:
Closes #289 

#289 describes a case when a valid metric name is sanitized (i.e., changed) when it need not be. I added a test to exercise this case and modified the `sanitize_metrics_name` function to pass this test. In terms of approach, I changed from using methods on strings to using a pre-constructed lookup table. The helper functions that were previously used to implement `sanitize_metrics_name` have been moved in the `test` module to act as oracles in the property tests.

### DIFF
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -226,7 +226,13 @@ mod tests {
 
     #[test]
     fn test_sanitize_metric_name_known_cases() {
-        let cases = &[("*", "_"), ("\"", "_"), ("foo_bar", "foo_bar"), ("1foobar", "_foobar")];
+        let cases = &[
+            ("*", "_"),
+            ("\"", "_"),
+            ("foo_bar", "foo_bar"),
+            ("foo1_bar", "foo1_bar"),
+            ("1foobar", "_foobar"),
+        ];
 
         for (input, expected) in cases {
             let result = sanitize_metric_name(input);
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -110,8 +110,17 @@ pub fn write_metric_line<T, T2>(
 /// [data model]: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
 pub fn sanitize_metric_name(name: &str) -> String {
     // The first character must be [a-zA-Z_:], and all subsequent characters must be [a-zA-Z0-9_:].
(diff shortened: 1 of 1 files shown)


tobz:
@gnuvince Thanks for opening the initial issue and the subsequent PR to fix it.

I think the lookup table approach is a bit too optimized.  I know, this might sound silly, but given the pattern of how an exporter is used -- typically on an interval or polled infrequently -- the lookup table-based approach feels like it's trading far too much in terms of obviousness and readability for what is likely a small improvement in overall performance.

I'd like to a see a more iterative, straightforward approach -- replace the initial character, even if it results in a suboptimal clone, then do the remaining steps -- as a first pass of fixing this issue.

gnuvince:
How about this version?

metrics-exporter-prometheus/src/formatting.rs
```diff
@@ -110,8 +110,17 @@ pub fn write_metric_line<T, T2>(
...
+    let mut is_invalid: fn(char) -> bool = invalid_metric_name_start_character;
+    for c in name.chars() {
+        if is_invalid(c) {
+            out.push('_');
+        } else {
+            out.push(c);
+        }
+        is_invalid = invalid_metric_name_character;
+    }
+    return out;
```
tobz, line 123: No need for the return here, as the last expression in scope will be used as the return value.

```suggestion
    out
```

tobz (changes requested):
Looking a lot simpler and straightforward, just one more small tweak.

tobz (approved):

tobz:
Released in `metrics-exporter-prometheus@v0.10.0`.

PR #296, opened 2022-04-01
fix prometheus metric name and label key sanitizer

wx-csy:
The original sanitizer mistakenly replaces digits in the middle of a string with an underscore. This pull request fixes this issue

### DIFF
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -341,13 +341,17 @@ mod tests {
 
             // Label keys cannot begin with two underscores, as that format is reserved for internal
             // use.
-            if as_chars.len() == 2 {
+            //
+            // TODO: More closely examine how official Prometheus client libraries handle label key sanitization
+            // and follow whatever they do, so it's not actually clear if transforming `__foo` to `___foo` would
+            // be valid, given that it still technically starts with two underscores.
+            /*if as_chars.len() == 2 {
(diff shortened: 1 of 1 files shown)


tobz:
Hey there, and thanks for your contribution!

I had actually just merged #290 today which fixed this problem for metric names, but not label keys.  Would you be willing to update this PR to utilize the same approach for label keys as was taken for metric names?

wx-csy:
> Hey there, and thanks for your contribution!
> 
> I had actually just merged #290 today which fixed this problem for metric names, but not label keys. Would you be willing to update this PR to utilize the same approach for label keys as was taken for metric names?

Updated. By the way, the original code adds a new underscore before keys starting with double underscores. But technically, triple underscores still start with double underscores. I think it'd better leave it as is.

tobz:
> Updated. By the way, the original code adds a new underscore before keys starting with double underscores. But technically, triple underscores still start with double underscores. I think it'd better leave it as is.

Hmmmmm, yeah... it's tricky because the regex they give for valid label keys is `[a-zA-Z_][a-zA-Z0-9_]*` but then they also say that label keys starting with `__` is reserved for internal use... so it's not clear if they only mean `__foo` or, if as you point out, `___foo` would be considered "reserved" because it happens to have two underscores at the front.

I agree that we can leave your PR as-is and I'll try to do some more research to see how officially libraries handle this, if they handle it at all.

tobz:
Released in `metrics-exporter-prometheus@v0.10.0`.

PR #302, opened 2022-05-23
Improve handling of the global recorder instance

Nilstrieb:
This gets rid of the dangerous `static mut`, adds more comments about the code, relaxes the orderings and documents the unsoundness of the `clear` function.

It implements a small once_cell-like abstraction.

I also noticed that the whole `#[cfg(atomic_cas)]` seems pretty pointless since this crate doesn't support `no_std` anyways.

About `clear_recorder`: This function was unsound before as well, since the following interleaving was possible:
- Thread A reads the state as init
- Thread B clears
- Thread B CASes the state and starts writing to the recorder
- Thread A now reads the recorder - a data race!


metrics/src/recorder.rs
```diff
@@ -1,13 +1,102 @@
...
+            }
+        }
+
+        pub unsafe fn set_racy(
+            &self,
+            recorder: &'static dyn Recorder,
+        ) -> Result<(), SetRecorderError> {
+            match self.state.load(Ordering::Acquire) {
+                UNINITIALIZED => {
+                    // SAFETY: Access is unique because we CASed the state to INITIALIZING above
```
5225225, line 77: This safety comment doesn't sound right, access is unique because the safety precondition of `set_racy` is that this isn't ran concurrently.

metrics/src/recorder.rs
```diff
@@ -1,13 +1,102 @@
...
+                    self.recorder.get().write(Some(recorder));
+                    self.state.store(INITIALIZED, Ordering::Release);
+                    Ok(())
+                }
+                INITIALIZED => {
```
5225225, line 82: is this meant to be INITIALIZING? It's fine if set_racy is called on an already init cell, just as long as it's not concurrent.

Nilstrieb: ah yes, it should, thanks!

metrics/src/recorder.rs
```diff
@@ -1,13 +1,102 @@
...
+    /// The recorder has been initialized successfully and can be read
+    const INITIALIZED: usize = 2;
+
+    impl RecorderOnceCell {
+        pub const fn new() -> Self {
+            Self { recorder: UnsafeCell::new(None), state: AtomicUsize::new(UNINITIALIZED) }
+        }
+
+        #[cfg(atomic_cas)]
+        pub fn set(&self, recorder: &'static dyn Recorder) -> Result<(), SetRecorderError> {
```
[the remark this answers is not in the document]
Nilstrieb: The caller is only ever supposed to call this once, so it's not contended.

metrics/src/recorder.rs
```diff
@@ -1,13 +1,105 @@
...
+    const UNINITIALIZED: usize = 0;
+    /// The recorder is currently being initialized
+    const INITIALIZING: usize = 1;
+    /// The recorder has been initialized successfully and can be read
+    const INITIALIZED: usize = 2;
+
+    impl RecorderOnceCell {
+        #[cfg(atomic_cas)]
+        pub fn set(&self, recorder: &'static dyn Recorder) -> Result<(), SetRecorderError> {
+            // Acquire the lock because the write below must not be reordered above the CAS
```
tobz, line 34:
```suggestion
            // Acquire the lock because the write below must not be reordered above the CAS.
```

metrics/src/recorder.rs
```diff
@@ -1,13 +1,105 @@
...
+        state: AtomicUsize::new(UNINITIALIZED),
+    };
+
+    /// The global Recorder instance with a `once_cell`-like API
+    pub struct RecorderOnceCell {
+        recorder: UnsafeCell<Option<&'static dyn Recorder>>,
+        state: AtomicUsize,
+    }
+
+    /// The recorder is uninit and can be set
```
tobz, line 24:
```suggestion
    /// The recorder is uninit and can be set.

File: metrics/src/recorder.rs [hunk 6]
```diff
@@ -1,13 +1,105 @@
...
+    const UNINITIALIZED: usize = 0;
+    /// The recorder is currently being initialized
```
tobz, line 26:
```suggestion
    /// The recorder is currently being initialized.

File: metrics/src/recorder.rs [hunk 7]
```diff
@@ -1,13 +1,105 @@
...
+    const INITIALIZING: usize = 1;
+    /// The recorder has been initialized successfully and can be read
```
tobz, line 28:
```suggestion
    /// The recorder has been initialized successfully and can be read.

File: metrics/src/recorder.rs [hunk 8]
```diff
@@ -1,13 +1,105 @@
...
+                    self.state.store(INITIALIZED, Ordering::Release);
+                    Ok(())
+                }
+                _ => Err(SetRecorderError(())),
+            }
+        }
+
+        /// # Safety
+        /// The caller must guarantee that no reader has read the state before we do this and then
+        /// reads the recorder after another writer has written to it after us.
```
tobz, line 56: This needs an actual high-level description, not just the safety section. I realize it's not actually part of the crate's public API, but if it's going to have documentation at all, it needs to be well-formed.

Something as simple as `/// Clears the currently installed recorder.` would suffice.

metrics/src/recorder.rs
```diff
@@ -168,9 +223,19 @@ pub unsafe fn set_recorder_racy(recorder: &'static dyn Recorder) -> Result<(), S
...
+/// # Warning
+///
+/// This function is unsound and must not be called during any readers reading or writers writing.
+/// The caller can cause readers and writers to race if they are in reading/writing while
+/// this function is called.
 #[doc(hidden)]
 pub fn clear_recorder() {
-    STATE.store(UNINITIALIZED, Ordering::SeqCst);
+    // SAFETY: We cannot guarantee this here :/
+    // FIXME: Fix this unsoundness - either have a full `RwLock` around `recorder`, or make this unsafe
```
tobz, line 235: I'd say we should just drop these internal comments and mark the overall function `unsafe`. The warning section covers the unsafety and lets users know what they have to do. We're never going to have a lock around the global recorder.

tobz (approved):
This is looking good to me, overall. Just a few nits related to documentation.

I'm still working on looking at the benchmarks because I was curious if the performance is identical for loading the recorder, given the slight difference between using `Option<T>` now vs having `NoopRecorder`.

metrics/src/recorder.rs
```diff
@@ -1,13 +1,105 @@
...
+mod cell {
+    use super::{Recorder, SetRecorderError};
+    use std::cell::UnsafeCell;
+    use std::sync::atomic::{AtomicUsize, Ordering};
 
-const UNINITIALIZED: usize = 0;
-const INITIALIZING: usize = 1;
-const INITIALIZED: usize = 2;
+    // FIXME: This can't be a const new function because trait objects aren't allowed in const fns
+    // This was stabilized in 1.61, so it can be cleaned up when it becomes the MSRV
```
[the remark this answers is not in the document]
tobz: Gotcha, I'm good with leaving the comment as-is, then. 👍🏻

metrics/src/recorder.rs
```diff
@@ -1,13 +1,105 @@
+use self::cell::RecorderOnceCell;
 use crate::{Counter, Gauge, Histogram, Key, KeyName, Unit};
 use core::fmt;
-use core::sync::atomic::{AtomicUsize, Ordering};
 
-static mut RECORDER: &'static dyn Recorder = &NoopRecorder;
```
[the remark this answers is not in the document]
Nilstrieb: It is still required as the default for `recorder`. The `Option<T>` is only used internally because `once_cell`  uses it as well. I can trivially change it back to also store the `NoopRecorder` in the cell instead of the option.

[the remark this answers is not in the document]
tobz: Ah yeah, true. I'll probably nuke `metrics::recorder()` to get rid of `NoopRecorder` but nothing that has to happen in this PR.

tobz:
Everything checked out performance-wise. 👍🏻 

Thank you for your contribution!

tobz:
Released in `metrics@v0.19.0` (along with other workspace crate releases to move to the new minor.)

PR #303, opened 2022-05-26
Fix `metrics::Cow` provenance issue

Nilstrieb:
Calling `.as_mut_ptr` on a `String` actually goes through `&mut str`, which shrinks the provenance of the pointer to only contain the initialized bytes. This caused issues when a reconstructed `String` tried to write to the uninitialized part of it. The fix is to go through `Vec::<u8>::as_mut_ptr`, which gives provenance for the entire allocation.

Also PRed upstream [beef#47](https://github.com/maciejhirsz/beef/pull/47)

### DIFF
--- a/metrics/src/cow.rs
+++ b/metrics/src/cow.rs
@@ -330,7 +330,9 @@ unsafe impl Cowable for str {
 
     #[inline]
     fn owned_into_parts(owned: String) -> (NonNull<u8>, Metadata) {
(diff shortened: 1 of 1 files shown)


tobz (approved):
Works for me.

Guess I should also run the tests under Miri at some point, too. :)

tobz:
Thanks for your contribution!

tobz:
Released in `metrics@v0.19.0`.

PR #304, opened 2022-05-27
Quantile Remapping Fix

dfwilbanks395:
This PR includes the following:

- Update to DDSketch 0.1.3 which fixes an issue mapping quantiles to rank.
- Updates Summary to return the min and max values when a request for the 0.0 or 1.0 quantile is made. This is convenient as described in 5274708.
- Resolves an issue with quantile mapping in Summary::quantile(), where the Summary could previously return the value that belongs to the adjacent rank. Also resolves incorrect assertions in a test, that now pass with the changes to quantile mapping.


tobz (approved):
Makes sense to me! 👍🏻 

Thanks for your contribution.

tobz:
Released in `metrics@v0.19.0`.

Issue #269, opened 2022-02-01
[metrics-exporter-prometheus] Summary histograms do not use sliding time windows


lilymara-onesignal:
According to the [prometheus docs](https://prometheus.io/docs/instrumenting/writing_clientlibs/#summary) summaries should be calculated based on sliding time windows. Looking at the code, this appears to not be implemented for [`metrics_util::Summary`](https://github.com/metrics-rs/metrics/blob/main/metrics-util/src/summary.rs), which is instead a summary over the entire range of data throughout the application's total runtime. This means that long-running applications will have their histogram values stabilize and spikes in the dataset will become virtually invisible. This is a graph of a metric with a previous internal monitoring tool which did use sliding time windows for summaries, which was switched over to metrics-exporter-prometheus around 22:00UTC on 2022-01-27. You can see the granularity of the data disappear after the switch and the data begin to drift towards a stable value. 

![Screen Shot 2022-02-01 at 11 07 55](https://user-images.githubusercontent.com/54288692/152034896-5a6d7910-3de1-48a8-b6ff-efed44ae8e98.png)

Either metrics-exporter-prometheus or [`metrics_util::Summary`](https://github.com/metrics-rs/metrics/blob/main/metrics-util/src/summary.rs) should provide by default or provide optionally, sliding time windows.

tobz:
Thanks for reporting this.

While I understand the current behavior of aggregated summaries is suboptimal for long-running processes, I'm not sure I have the time to craft an enhancement to make them use a sliding time window under the hood.

I'd be willing to review a PR that does as much, as well as gently mentor/guide such a PR.

tobz:
Very belated update, but: this should be solved as of #306, which was released as `metrics-exporter-prometheus@v0.11.0`.

PR #306, opened 2022-06-09
Add RollingSummary to prevent summary saturation

danielnelson:
This PR attempts to address the issue described in #269 where Prometheus
Summaries become saturated.  I added a RollingSummary type that manages a list
of Summary so that old values can be retired.

There is a breaking change in `AtomicStorage`:
```diff
- type Histogram = Arc<AtomicBucket<f64>>;
+ type Histogram = Arc<AtomicBucket<(f64, Instant)>>;
```

I tested using a program that produces a test pattern, every two minutes it
switches between uniform random values between 0-100 and 100-200.

I also graph the summary `rate(sum[1m]) / rate(count[1m])` for comparison to
the 50th quantile.

Before this patch:

2022-06-09-163831_1347x637_scrot

With this patch:

2022-06-09-160026_1345x635_scrot

### DIFF
--- a/metrics-exporter-prometheus/src/distribution.rs
+++ b/metrics-exporter-prometheus/src/distribution.rs
@@ -1,4 +1,8 @@
+use std::num::NonZeroU32;
+use std::time::Duration;
 use std::{collections::HashMap, sync::Arc};
+
+use quanta::Instant;
 
 use crate::common::Matcher;
 
@@ -18,7 +22,7 @@ pub enum Distribution {
     /// Computes and exposes value quantiles directly to Prometheus i.e. 50% of
     /// requests were faster than 200ms, and 99% of requests were faster than
     /// 1000ms, etc.
-    Summary(Summary, Arc<Vec<Quantile>>, f64),
+    Summary(RollingSummary, Arc<Vec<Quantile>>, f64),
 }
 
 impl Distribution {
@@ -30,17 +34,19 @@ impl Distribution {
 
     /// Creates a summary distribution.
     pub fn new_summary(quantiles: Arc<Vec<Quantile>>) -> Distribution {
-        let summary = Summary::with_defaults();
+        let summary = RollingSummary::default();
         Distribution::Summary(summary, quantiles, 0.0)
     }
 
     /// Records the given `samples` in the current distribution.
-    pub fn record_samples(&mut self, samples: &[f64]) {
+    pub fn record_samples(&mut self, samples: &[(f64, Instant)]) {
         match self {
-            Distribution::Histogram(hist) => hist.record_many(samples),
+            Distribution::Histogram(hist) => {
+                hist.record_many(samples.iter().map(|(sample, _ts)| sample));
+            }
             Distribution::Summary(hist, _, sum) => {
-                for sample in samples {
-                    hist.add(*sample);
+                for (sample, ts) in samples {
+                    hist.add(*sample, *ts);
                     *sum += *sample;
                 }
             }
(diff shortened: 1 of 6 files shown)


danielnelson:
In a staging environment I have ran into this error:
```
panicked at ‘index out of bounds: the len is 10939 but the index is 10939’, /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/sketches-ddsketch-0.1.3/src/store.rs:155:22
```

I'll try to find a simple way to reproduce.

tobz:
First off, thank you for submitting a solution for this problem. :)

I'll need to see what the benchmarks look like, but the tracking of a timestamp per recorded value in a histogram is a bit of a red flag to me, performance-wise.

There's some thoughts I have in mind for how to avoid it while still providing proper time bucketing/rolling, but I'll wait to see what the benchmarks look like first before getting into any refactoring.

tobz:
Alright, apologies for the delay.

After looking at the change, and running `metrics-benchmark` this adds more overhead than I'm comfortable with. Throughput seems to drop by around ~15%, which means it's actually _higher_ because the benchmark app basically does a loop of "counter, gauge, histogram", so if 1/3" of it is contributing an overall ~15% slowdown.... but anyways, this is also in conjunction with the fact it's forcing everyone else using `Handle` to have to upgrade in the future and pay that cost, even if they don't need to.

I think what I'd want to see is localizing the change to `metrics-exporter-prometheus`. This should be doable without out much additional effort by creating a new `Storage` implementation in `metrics-exporter-prometheus` that emulates what you did to `AtomicStorage<K>`. You'd similarly have to create something like a newtype wrapper around `AtomicBucket<(f64, Instant)>` and impl `HistogramFn`.. but it should be doable.

As I try and get closer and closer to what a 1.0 release might look like, I'm striving to avoid continually changing every crate in the workspace due to changes that primarily occur in only one of them.

danielnelson:
That makes sense, I will try to make the suggested changes this week.

danielnelson:
I have updated the PR as suggested, please let me know what you think.  BTW, the issue linked above with ddsketch is not yet resolved and will need to be before this could be merged.

tobz:
Taking a cursory look, this a lot more in line with what I was thinking. 👍🏻 

There's probably some performance to be squeezed out if we lowered the timing accuracy by bucketing with a background task on an interval, rather than getting the timestamp at record time, but I don't think this matters a ton for users using the Prometheus exporter anyways.

I'll do a more in-depth path, including looking at the DDSketch issue you mentioned, in the next few days. Thank you for making the requested changes.

tobz:
Bah, looks like `indexmap` is unhappy on Rust 1.55. I'll get the MSRV bumped to 1.56 which should fix that and then we can see what CI looks like for this.

tobz:
I updated the CI workflow to bump the MSRV to 1.56.1 which seems to get the MSRV-specific unit tests passing again, so I think you should be good if you merge in `main`.

tobz (approved):
Was finally able to take a look through, and this looks solid to me.

Most of my thoughts/concerns are nits or can be addressed in the future (i.e. configurable summary window/bucket width, enforcing monotonicity of samples to drop a lot of the logic in `add`, etc) and nothing here feels like a design pitfall.

tobz:
Thanks for your contribution! <3

tobz:
Released as `metrics-exporter-prometheus@v0.11.0`.

tobz:
Released as `metrics@v0.20.0`.

Issue #287, opened 2022-03-16
Bump MSRV to 1.56


rukai:
Although there isnt an MSRV specified in the readme or elsewhere the CI does specify runs on rust 1.55 so I assume there is an undocumented MSRV of 1.55?

Bumping to 1.56 will allow updating to hashbrown 0.12.0

The old version of hashbrown is causing duplicate hashbrown dependencies in my dependency tree.

tobz:
Woops, kind of forgot about this one.

I recently bumped the MSRV in CI to 1.56.1, although I haven't yet codified it via `rust-version` in the various Cargo.toml files.

PR #311, opened 2022-07-05
Document CI enforced msrv in readme and rust-version fields

rukai:
closes https://github.com/metrics-rs/metrics/issues/287

If there is a policy of when the MSRV can be bumped I would love to document that in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,11 @@ As well, there are also some community-maintained exporters and other integratio
 * [`metrics-exporter-newrelic`][metrics-exporter-newrelic]: A `metrics`-compatible exporter for sending metrics to New Relic.
 * [`opinionated_metrics`][opinionated-metrics]: Opinionated interface to emitting metrics for CLi/server applications, based on `metrics`.
 
+## MSRV
+
+Minimum Supported Rust version is **1.56.1**
+It is enforced in CI.
+
 # contributing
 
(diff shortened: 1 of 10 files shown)


tobz:
I think, conceptually, I'd be OK with the policy being:

- goal is to try and support older versions where possible (not opting in to newer versions just to use a new helper method on standard library types, etc)
- for any bump in MSRV, we must support at least the last 4 stable releases
- we do not bump the MSRV for newer versions of dependencies in core crates (`metrics` and `metrics-util`)

For the third point, this will probably end up needing a slight amount of nuance, because `metrics-util` touches a lot of stuff... especially depending on feature flags. As an example, though: I really dislike having to bump the MSRV just to pull in a newer version of `hashbrown`, or something to that effect, because slimming transitive dependencies strikes me as a fool's errand, personally. Keeping a wider range of supported versions is a higher priority to me. If we needed to bump `hashbrown` to get access to a higher performance table API, though, for example, and that necessitated a bump in MSRV... that's probably a case where we could agree it was worth it.

rukai:
Alright, I added in the bullet points with some minor simplifications.

tobz (approved):

tobz:
Works for me. I might do some stylistic tweaks later but better to get the MSRV documented sooner rather than later.

Thanks for your contribution!

Issue #310, opened 2022-06-28
Describe metrics with owned strings


fredr:
I'm working on a project that reads metric descriptions from a [config file](https://github.com/fredr/data-exporter/blob/main/examples/config.yaml#L3). But since the description needs to be a `&'static str`, I can't use the `describe_*` macros.

Is there a way of getting around that, or would it be possible to allow the description to be an owned string (like it is with the name)?

PR #312, opened 2022-07-05
Change description to be `SharedString`

fredr:
This makes it possible to describe metrics using both `&'static str` as well as from an owned string.

Solves #310

### DIFF
--- a/metrics-benchmark/src/main.rs
+++ b/metrics-benchmark/src/main.rs
@@ -4,7 +4,7 @@ use hdrhistogram::Histogram as HdrHistogram;
 use log::{error, info};
 use metrics::{
     gauge, histogram, increment_counter, register_counter, register_gauge, register_histogram,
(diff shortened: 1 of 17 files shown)


tobz:
Ahhhh, nice. It hadn't yet struck me that since the work was already there to allow constants for descriptions, that it could already be an owned string so long as we had a way to convert it.

fredr:
I just trialed and errored my way through, wasn't sure if this was the way you wanted to solve it. But it seems to work well!

tobz:
Yeah, no, it's a totally fine solution. Even though `SharedString` is in the public API already, I might end up wrapping this up later on in some fashion in the same way that `Key` and `Label` are wrapped... but not an issue that needs to be addressed in this PR.

fredr:
Cool! I started by re-using `KeyName`, but it felt wrong to use that type for this

tobz:
Thanks for your contribution!

tobz:
Released as `metrics@v0.20.0`.

Issue #235, opened 2021-10-15
Support for 32-bit architectures?


jszwedko:
It looks like currently this crate doesn't build on `powerpc-unknown-linux-gnu`:

```
➜  metrics git:(main) cargo build --target=powerpc-unknown-linux-gnu
   Compiling serde v1.0.130
   Compiling prost v0.7.0
   Compiling tracing v0.1.29
   Compiling thiserror v1.0.30
   Compiling metrics v0.17.0 (/home/jesse/workspace/metrics/metrics)
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> metrics/src/key.rs:7:32
  |
7 |     sync::atomic::{AtomicBool, AtomicU64, Ordering},
  |                                ^^^^^^^^^
  |                                |
  |                                no `AtomicU64` in `sync::atomic`
  |                                help: a similar name exists in the module: `AtomicU8`

error: aborting due to previous error
```

We discovered this by way of a user [trying to compile vector](https://github.com/vectordotdev/vector/discussions/9560) for `powerpc-unknown-linux-gnu`.

Is this a platform this crate would be interested in supporting? I imagine this would mean generally supporting 32-bit architectures.

tobz:
No issue with trying to support them, nope!

Likely just need to do something like [this](https://github.com/metrics-rs/quanta/pull/52/files) across the entire crate.

PR #313, opened 2022-07-11
Shims remaining AtomicU64 usage

huntc:
Shims the use of `AtomicU64` in the main metrics lib so that we can use the library on MIPS and other 32 bit platforms.

Fixes https://github.com/metrics-rs/metrics/issues/235


huntc:
Needs some more work... some of the atomic methods being used aren't available on MIPS either...

huntc:
The problem I'm faced with here is that `fetch_max` and `fetch_update` do not exist in the shimmed AtomicU64. I've created an [issue there](https://github.com/bltavares/atomic-shim/issues/3) and will attempt a PR later in the week. Appreciate any other thinking on this too though. Thanks.

huntc:
One other thought: could we get away with 32 bit has values for keys?

tobz:
> One other thought: could we get away with 32 bit hash values for keys?

We cannot, no: we need to be able to hold a `u64` in order to satisfy `std::hash::Hasher::finish`.

huntc:
> > One other thought: could we get away with 32 bit hash values for keys?
> 
> We cannot, no: we need to be able to hold a `u64` in order to satisfy `std::hash::Hasher::finish`.

Could a 32 bit hash be used as an opt-in feature, or even more generally? We then cast the u64 to u32 for the atomic.

tobz:
> Could a 32 bit hash be used as an opt-in feature, or even more generally? We then cast the u64 to u32 for the atomic.

Doing so would imply that we need to drop entropy from the original 64-bit hash, which I don't want to do. Performance is a top priority of `metrics`, and adding a simple way to effectively/silently truncate the hash values for keys, possibly leading to an increased rate of hash collisions, doesn't seem worth it to me as a shortcut to make this change.

Another crate to try out might be `portable-atomic`: it's a bit more generalized, has fewer transitive dependencies than `atomic-shim`, and most importantly, should have support for `AtomicU64::fetch_max` and `AtomicU64::fetch_update`.

huntc:
> Another crate to try out might be portable-atomic: it's a bit more generalized, has fewer transitive dependencies than atomic-shim, and most importantly, should have support for AtomicU64::fetch_max and AtomicU64::fetch_update.

Thanks. I’ll take a look in the next few days.

huntc:
> Another crate to try out might be `portable-atomic`: it's a bit more generalized, has fewer transitive dependencies than `atomic-shim`, and most importantly, should have support for `AtomicU64::fetch_max` and `AtomicU64::fetch_update`.

I've now updated my PR to use `portable-atomic`. All appears well given a `target.mipsel-unknown-linux-musl` build.

tobz (approved):

tobz:
Sorry, been a bit preoccupied lately.

Looking it over, I'm happier with this yes. I'll merge it once CI passes.

huntc:
> Sorry, been a bit preoccupied lately.
> 
> Looking it over: I'm happier with this, yes. I'll merge it once CI passes.

Thanks! I'm unsure why that target build is failing though... any ideas?

huntc:
> 

FYI I tried building the `x86_64-unknown-linux-gnu` target locally on `metrics-tracing-context` and all was well...

tobz:
Thanks for your contribution!

huntc:
> Thanks for your contribution!

My pleasure. Thank you for your massive contribution with this library also!

PR #316, opened 2022-07-27
docs: Remove incorrect return info

fredr:
Documentation states that the previous values are returned, but they are not.


tobz (approved):
Yep, good catch. 👍🏻

tobz:
Thanks again for your contribution. :)

Issue #175, opened 2021-02-09
Exemplar support


clux:
A way to attach traceids to histograms as per the [OpenMetics specifictaion](https://github.com/OpenObservability/OpenMetrics/blob/master/specification/OpenMetrics.md#exemplars-1)

Use case is [grafana tempo](https://grafana.com/go/webinar/getting-started-with-tracing-and-grafana-tempo/)

tobz:
Taking a cursory glance, this is definitely more work than I'd be comfortable taking on personally.  I'd accept a PR adding this functionality and can try and offer some guidance on working through it if need be.

I'll leave the issue open for anyone else looking for the same thing.

clux:
Completely understand. Mostly wanted to see if there was interest around for it as it looked cool. Thanks for keeping a tracking issue.

tobz:
Nobody is working on this as far as I know, so it's all yours. :)

wperron:
@gsserge any progress so far? I'm interested in this feature, I'd be glad to help get it through

fredr:
@gsserge @wperron did you start working on this? I'm thinking of looking into this in the upcoming weeks, so if you have made any progress or figured something out while looking into or have any other info regarding this it would be appreciated

wperron:
> @gsserge @wperron did you start working on this? I'm thinking of looking into this in the upcoming weeks, so if you have made any progress or figured something out while looking into or have any other info regarding this it would be appreciated

Hey 👋 I have not worked on this no, changing priorities and all, feel free to take it

fredr:
hey @campbellC, I'm happy to continue with it, but waiting for some feedback on how it should be implemented. 

The PR so far is just a proof of concept and a base for discussion, and I've dropped the momentum a bit since the summer, but I will try to get it rolling again.

If you have any feedback on the PR or ideas how it could be implemented differently, let me know!

PR #320, opened 2022-08-02
Add openmetrics exemplar support

fredr:
This is a work in progress implementation for #175 

Opening this as a draft to have a discussion to see if this is a valid approach to go forward with, or if there is a different path that is cleaner/better.

I've tried to keep all changes to the prometheus exporter, as this is a prometheus-specific feature, but have had to add downcasting to the recorder, so that a handle to the specific recorder can be fetched. But maybe there is some more generic api that can be used. I'm not very familiar with other metrics systems, but maybe it is common to add additional data to observations?

There is an example of how to increase a counter and record a sample in a histogram with exemplars. This is extra problematic for histograms, as we need to know what bucket to assign the exemplar with.

I'm also thinking about what would be a good macro api for this. Since an exemplar is just an other set of labels, we can't just add it as an other parameter.

e.g. with the following its hard to know what labels should be metric labels and which should be exemplar labels 
```
counter_with_exemplar!("my_counter", "some_label" => "value", "trace_id" => "123");
```
We could probably say that exemplar labels are always expressions, so something like would probably be possible (I haven't written any macros, but I'm guessing this is possilbe)
```
counter_with_exemplar!("my_counter", vec![Label::from_parts("trace_id", "123")], "some_label" => "value");
counter_with_exemplar!("my_counter_without_labels", vec![Label::from_parts("trace_id", "123")]);
```
It might be confusing still, as it is easy to mix the different labels up.

Also, where would these macros live? I don't think we can add them to metrics-macros (behind a feature), since that would be a cyclic dependency if it needed to depend on the PrometheusRecorder.

### DIFF
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -51,6 +51,10 @@ required-features = ["push-gateway"]
 name = "prometheus_server"
 required-features = ["http-listener"]
 
+[[example]]
+name = "exemplars"
+required-features = ["http-listener"]
+
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
--- a/metrics-exporter-prometheus/examples/exemplars.rs
+++ b/metrics-exporter-prometheus/examples/exemplars.rs
@@ -1,0 +1,100 @@
+use std::thread;
+use std::time::Duration;
+
+use metrics::{
+    decrement_gauge, describe_counter, describe_histogram, gauge, histogram, increment_counter,
+    increment_gauge, Key, Label,
+};
+use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusRecorder};
+use metrics_util::MetricKindMask;
+
+use quanta::Clock;
+use rand::{thread_rng, Rng};
+
+fn main() {
+    tracing_subscriber::fmt::init();
+
+    let builder = PrometheusBuilder::new();
+    builder
+        .idle_timeout(
+            MetricKindMask::COUNTER | MetricKindMask::HISTOGRAM,
+            Some(Duration::from_secs(10)),
+        )
+        .set_buckets(&[0.01, 0.1, 0.2, 0.5, 1., 2.])
+        .unwrap()
+        .install()
+        .expect("failed to install Prometheus recorder");
+
+    // We register these metrics, which gives us a chance to specify a description for them.  The
+    // Prometheus exporter records this description and adds it as HELP text when the endpoint is
+    // scraped.
+    //
+    // Registering metrics ahead of using them is not required, but is the only way to specify the
+    // description of a metric.
+    describe_counter!("tcp_server_loops", "The iterations of the TCP server event loop so far.");
+    describe_histogram!(
+        "tcp_server_loop_delta_secs",
+        "The time taken for iterations of the TCP server event loop."
+    );
+
+    let clock = Clock::new();
+    let mut last = None;
+
+    increment_counter!("idle_metric");
+    gauge!("testing", 42.0);
+
+    // Loop over and over, pretending to do some work.
+    loop {
+        let with_exemplar = thread_rng().gen_bool(0.75);
+        if with_exemplar {
+            let labels = vec![Label::new("system", "foo")];
+            let metric_key = Key::from_parts("tcp_server_loops", labels);
+            let recorder = metrics::recorder()
+                .downcast_ref::<PrometheusRecorder>()
+                .expect("Couldn't downcast to prometheus recorder");
+            let handler = recorder.register_counter_with_exemplar(&metric_key);
+            handler.increment_with_exemplar(1, vec![Label::new("trace_id", "xyasfj234234")]);
+        } else {
+            increment_counter!("tcp_server_loops", "system" => "foo");
+        }
+
+        if let Some(t) = last {
+            let delta: Duration = clock.now() - t;
+            if with_exemplar {
+                let labels = vec![Label::new("system", "foo")];
+                let metric_key = Key::from_parts("tcp_server_loops", labels);
+                let recorder = metrics::recorder()
+                    .downcast_ref::<PrometheusRecorder>()
+                    .expect("Couldn't downcast to prometheus recorder");
+                let distribution = recorder.get_distribution(metric_key.name());
+                let handler = recorder.register_histogram_with_exemplar(&metric_key);
+                handler.record_with_exemplar(
+                    distribution,
+                    metrics::__into_f64(delta),
+                    vec![Label::new("trace_id", "xyasfj234234")],
+                );
+            } else {
+                histogram!("tcp_server_loop_delta_secs", delta, "system" => "foo");
+            }
+        }
+
+        let increment_gauge = thread_rng().gen_bool(0.75);
+        if increment_gauge {
+            increment_gauge!("lucky_iterations", 1.0);
+        } else {
+            decrement_gauge!("lucky_iterations", 1.0);
+        }
+
+        last = Some(clock.now());
+
+        let render = thread_rng().gen_bool(0.1);
+        if render {
+            let recorder = metrics::recorder()
+                .downcast_ref::<PrometheusRecorder>()
+                .expect("Couldn't downcast to prometheus recorder");
+            println!("{}", recorder.handle().render());
+        }
+
+        thread::sleep(Duration::from_millis(750));
+    }
+}
--- a/metrics-exporter-prometheus/src/common.rs
+++ b/metrics-exporter-prometheus/src/common.rs
@@ -3,6 +3,7 @@ use std::collections::HashMap;
 use crate::distribution::Distribution;
 
 use crate::formatting::sanitize_metric_name;
+use crate::registry::Exemplar;
 use indexmap::IndexMap;
 use metrics::SetRecorderError;
 use thiserror::Error;
@@ -77,7 +78,8 @@ pub enum BuildError {
 }
 
 pub struct Snapshot {
-    pub counters: HashMap<String, HashMap<Vec<String>, u64>>,
-    pub gauges: HashMap<String, HashMap<Vec<String>, f64>>,
-    pub distributions: HashMap<String, IndexMap<Vec<String>, Distribution>>,
+    pub counters: HashMap<String, HashMap<Vec<String>, (u64, Option<Exemplar<u64>>)>>,
+    pub gauges: HashMap<String, HashMap<Vec<String>, (f64, Option<Exemplar<f64>>)>>,
+    pub distributions:
+        HashMap<String, IndexMap<Vec<String>, (Distribution, HashMap<usize, Exemplar<f64>>)>>,
 }
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -1,7 +1,11 @@
 //! Helpers for rendering metrics in the Prometheus exposition format.
 
+use std::time::SystemTime;
+
 use indexmap::IndexMap;
 use metrics::Key;
+
+use crate::registry::Exemplar;
 
 /// Breaks a key into the name and label components, with optional default labels.
 ///
@@ -57,16 +61,18 @@ pub fn write_type_line(buffer: &mut String, name: &str, metric_type: &str) {
 /// for aggregated histograms, or `quantile` for aggregated summaries.
 ///
 /// [exposition format]: https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-format-details
-pub fn write_metric_line<T, T2>(
+pub fn write_metric_line<T, T2, E>(
     buffer: &mut String,
     name: &str,
     suffix: Option<&'static str>,
     labels: &[String],
     additional_label: Option<(&'static str, T)>,
     value: T2,
+    exemplar: Option<Exemplar<E>>,
 ) where
     T: std::fmt::Display,
     T2: std::fmt::Display,
+    E: std::fmt::Display,
 {
     buffer.push_str(name);
     if let Some(suffix) = suffix {
@@ -102,6 +108,27 @@ pub fn write_metric_line<T, T2>(
 
     buffer.push(' ');
     buffer.push_str(value.to_string().as_str());
+
+    if let Some(exemplar) = exemplar {
+        buffer.push_str(" # ");
+        buffer.push('{');
+        for label in exemplar.labels {
+            buffer.push_str(label.key());
+            buffer.push_str("=\"");
+            buffer.push_str(label.value());
+            buffer.push('"');
+        }
+        buffer.push('}');
+
+        buffer.push(' ');
+        buffer.push_str(exemplar.value.to_string().as_str());
+
+        if let Ok(ts) = exemplar.timestamp.duration_since(SystemTime::UNIX_EPOCH) {
+            buffer.push(' ');
+            buffer.push_str(ts.as_secs_f64().to_string().as_str());
+        }
+    }
+
     buffer.push('\n');
 }
 
--- a/metrics-exporter-prometheus/src/lib.rs
+++ b/metrics-exporter-prometheus/src/lib.rs
@@ -115,3 +115,7 @@ mod recorder;
 mod registry;
 
 pub use self::recorder::{PrometheusHandle, PrometheusRecorder};
+
+pub use self::registry::Counter;
+pub use self::registry::Gauge;
+pub use self::registry::Histogram;
--- a/metrics-exporter-prometheus/src/recorder.rs
+++ b/metrics-exporter-prometheus/src/recorder.rs
@@ -13,12 +13,14 @@ use crate::distribution::{Distribution, DistributionBuilder};
 use crate::formatting::{
     key_to_parts, sanitize_metric_name, write_help_line, write_metric_line, write_type_line,
 };
-use crate::registry::GenerationalAtomicStorage;
+use crate::registry::{Exemplar, GenerationalAtomicStorage};
 
 pub(crate) struct Inner {
     pub registry: Registry<Key, GenerationalAtomicStorage>,
     pub recency: Recency<Key>,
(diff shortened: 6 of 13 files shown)


tobz:
Is there anyway you can give me a high-level overview of exemplars and how a typical application, with one of the official Prometheus client SDKs, is using them?

I'm finding it hard to conceptualize, and based on the changes so far, I think we need to step back and talk about the feature first before trying to design the interface to it.

fredr:
> Is there anyway you can give me a high-level overview of exemplars and how a typical application, with one of the official Prometheus client SDKs, is using them?

Absolutely, I'll write some stuff that I know about it here, let me know if there is any other details that you would like to know.

### Use case

Exemplars is a prometheus invention that can be use to attach additional data to some samples for a metric. An exemplar consist of the recorded value, a timestamp when it happen, and additional labels, the timestamp is optional (havent tried what happens if it is not set, guessing that the position when rendering it might get the collection time and not exactly when it happen). This is then used in e.g grafana to show little dots in graphs when creating graphs over these metrics. When hovering these dots the additional labels are shown. The most common use for this is to add a trace id to the exemplar labels, so that grafana can add a link into the tracing system.

This is one of the main trace discoverability features when using tempo as a tracing backend.

An example screenshot of how it is displayed in grafana:
image

### Exporting

For prometheus to be able to ingest exemplars, they need to be exported with the metric. Each label set for a metric can expose one exemplar. So in prometheus there will at most be one exemplar per metric + label set per collection. If multiple exemplars are written to the same metric + label set within the same collection, it will just overwrite the exemplar. (I would say that exemplars make most sense when using prometheus histograms where there can be one exemplar per bucket, since the bucket is also a label)

The format it is exposed in is defined in [open metrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#exemplars-1).

### Other implementations

I've mostly looked at how this is implemented in the prometheus golang client. E.g [how it is used on histograms](https://github.com/prometheus/client_golang/blob/66837e3298bdc57a828794c23bacb253ed8c04cd/prometheus/histogram.go#L307-L315). It is backed by a [atomic value store](https://github.com/prometheus/client_golang/blob/66837e3298bdc57a828794c23bacb253ed8c04cd/prometheus/histogram.go#L436) that just gets overwritten. The value that is stored is found [here](https://github.com/prometheus/client_model/blob/master/go/metrics.pb.go#L457-L464)

### Other thoughts

I noticed now that exemplars should not be implemented for summaries and gauges, only for histograms and counters, and for counters the openmetrics documentation is a big vague. IMO it would be enough to implement only for histograms, but that is still the hardest one so we might as well implement it for both.

I'm not sure what happens if a metric doesn't get their exemplar updated and gets collected several times. I think prometheus handles that, but I need to do some more research.

tobz:
Just a note: this is still on my backlog to review, things have just been a bit hectic for me lately. 😅

fredr:
No worries and no stress, but thanks for letting me know!

I'm also on discord most of the time if you want to discuss something when you look at this

tobz:
So, my main question after reading your explanation (thank you for that!): do the exemplars have to be logically related?

Which is to say... if I have two concurrent tasks/threads/whatever emitting metrics and both of them hit the random number generator lottery of "you should track these metrics as exemplars", do all the metrics they touch need to have the same examplar label (trace ID, etc) or could some of the metrics have trace ID 1, and some have trace ID 2, etc?

fredr:
They can be different, there is no connection between exemplars over different metrics.

tobz:
> They can be different, there is no connection between exemplars over different metrics.

Alright, that's good news. 👍🏻 

Depending on the behavior necessary, it seems like it could be possible to get away with sampling a value at the point of actually rendering the metrics. That is, every time the metrics are rendered -- which is just when we get a scrape request, or our interval to hit the push gateway ticks -- we collect the outstanding histogram samples and pick one of them to be our new exemplar.

Avoiding new exporter-specific methods seems like the highest priority item in my mind. We should ideally be able to just collect exemplars with people using the `metrics` macros and the Prometheus exporter the same way they always have.

fredr:
Sorry for being so slow to reply. 

Yes, getting the exemplar on render would definitely be enough, and it sound like you are on to something smart, but I don't understand how exactly :sweat_smile:. How would one add an exemplar, if there where no exporter specific method for it?

tobz:
Right, so my thought is that the render logic would essentially be responsible for figuring out if it was time to sample a new exemplar for each unique histogram.

So we'd have the exemplar value itself, and probably a timestamp for "when was this exemplar observed?". In the render logic, where it checks to see if it needs to consume any more raw samples from the underlying histogram storage, we'd see how long ago we last captured an exemplar for the given histogram. If we've exceeded our timeout, then we take one of the samples we just consumed and make it our exemplar, and update our timestamp.

Explained more contextually:

We start with an exporter that has a default initial state: no metrics observed yet, etc. We'll refer to the time that these actions/operations occur with the `t=..` notation, denoting the time in seconds.

1. At `t=0`, `render` is called, and we consume all histograms from the registry, and discover a new one, `histogram_a`.
2. `histogram_a` has 10 samples (10 is just a random number, doesn't matter if it's 1 or 10000) and since we have not seen this histogram before, our "it's time to capture an exemplar" logic kicks in, and we randomly select one of those 10 samples as the exemplar value.
3. At `t=2`, `render` is called again, and we get 5 more samples for `histogram_a` when we consume it from the registry. For the sake of explanation, let's pretend our exemplar selection logic doesn't select a new exemplar unless it's been over 5 seconds since the current one was selected. Since only two seconds have elapsed, we don't choose a new exemplar.
4. At `t=6`, `render` is called again, and we get 7 more samples for `histogram_a` when we consume it from the registry`. Since we last selected an exemplar for this histogram at `t=0`, our exemplar selection logic now kicks in, and we randomly select one of the 7 samples we just consumed and use that as the new exemplar for the histogram.

So, overall, I'm forcing some design decisions here for the sake of explaining my idea:

- exemplars are randomly selected from the samples we consume for a histogram
- exemplars are only refreshed if enough time has elapsed

fredr:
But how do we register the exemplar labels with the value? they need to be added when recording the value, as they contain data that is not know by the metric itself, most commonly a trace id that refers to some external tracing system.

That trace id will be different for every value recorded (as long as there are not multiple metrics recorded during the same request, which I thought your initial question about)

tobz:
Are the labels for exemplars only meant for exemplars period? Like would you not typically include, say, a trade ID label _unless_ you wanted it to be an exemplar?

fredr:
Yes exactly, its a separate label set that the regular label set for the metric itself. The common use case is to attach a trace id to a specific observation, and what I mean is a distributed trace id that comes from outside, like from jaeger, tempo, zipkin etc.

Usually when you build some kind of web server with incoming http requests, there will be some ingress gateway that initiates a trace, creates a trace id and attaches it to the http request as headers, and then your web server reads these and attaches a bunch of "spans" to it, and upload those to whatever tracing backend you have, a trace id can span over multiple web servers, if they do requests to each other within the same initial request.

The exemplar is then used to find a trace in the tracing backend from a graph built by metrics. Usually on latency histograms, so that when you see your latency graph and want to figure out why some requests take a long time, you click on the little exemplar dot representing a specific observed request that took a long time to process.

Using the golang client, the definition would look something like this, specifying the metrics labels and buckets
```go
var (
	histogram = prometheus.NewHistogramVec(
		prometheus.HistogramOpts{
			Name:    "foo_latency",
			Buckets: []float64{0.01, 0.1, 1, 10},
		},
		[]string{"path"}
	)
)
```

Then when observing a value, it would be something like this, very simplified, but getting the trace id from the request, and adding it to the exemplar label-set.
```go
histogram.With(prometheus.Labels{"path": "/foo"}).(prometheus.ExemplarObserver).ObserveWithExemplar(
	latency, prometheus.Labels{"traceID": req.Header.Get("x-trace-id")},
)
```

This is what the prometheus exporter will render, just to clarify what it looks like:
```
# TYPE foo_latency histogram
foo_latency_bucket{path="/bar", le="0.01"} 0
foo_latency_bucket{path="/bar", le="0.1"} 8
foo_latency_bucket{path="/bar", le="1"} 11 # {trace_id="xyz123"} 0.67 1520879912.512
foo_latency_bucket{path="/bar", le="10"} 17 # {trace_id="abc123"} 9.8 1520879607.789
foo_latency_bucket{path="/bar", le="+Inf"} 17
```

In this example there are two different requests that have been recorded with an exemplar, one in bucket le=1 and one in bucket le=10, with their exact value, when they where observed, and this extra set of labels that include their trace id. (there can maximum be one observed exemplar per metric + metric labelset + bucket, if two are observed for the same, the later one just replaces the first one)

Sorry if I'm poorly trying to explain things that you already know, but just trying to explain the flow of how these things are used in the kind of area that I work with.

fredr:
Hey @tobz, I lost the momentum a bit here, but would appreciate some feedback. I'm happy to continue on this if I know what way to go.

tobz:
Hey @fredr! I'm no stranger to losing momentum. :)

I've started getting more serious about planning out the remaining work to bring `metrics` to a 1.0 release, and the general theme of that work is simplifying the public API surface as much possible: avoiding niche methods/functions, allowing for more flexible inputs, to let `metrics` adapt in the future.

With that said, I think my biggest concern, sort of right from the beginning: it doesn't feel right for there to be macros/methods/etc that are specific to Prometheus. I'm certainly not against third-party exporters having their own specific macros or something of the sort... but it's not the design pattern I want to promote in `metrics` or any of the official `metrics` ecosystem crates.

What I would want to see is figuring out a way to make the determination of when exemplars should be tracked something that functions more like a scoped behavior i.e. a function that takes a closure, and changes a thread-local to influence the behavior of any code running the closure. Alternatively, and maybe I'm still not entirely understanding how exemplars are typically tracked/triggered, but some sort of approach that was deterministic i.e. for any metric that has a specific label key, which is configured as part of the exporter itself, sample updates to metrics with that label at a configurable rate to derive exemplars.

With an approach as described above, it maintains one of the original design goals of `metrics`: if you're using a library, in your application, that is instrumented with `metrics`, you get that metric data "for free". It could be useful to include the metrics emitted by dependencies as part of an exemplar, but you can't do that if they need to use a specific macro.

I'm happy to continue providing feedback on the design as long as it's along the lines as described above, because that represents both the most ergonomic path, and easiest to maintain path, in my eyes. If executing on what I've posted above feels like it would consume too much time, I totally understand.

fredr:
> I've started getting more serious about planning out the remaining work to bring metrics to a 1.0 release, and the general theme of that work is simplifying the public API surface as much possible: avoiding niche methods/functions, allowing for more flexible inputs, to let metrics adapt in the future.

That sounds great, do you know already if that will have any implications on how buckets are registered for prometheus? or any other upcoming prometheus related changes?

> With an approach as described above, it maintains one of the original design goals of metrics: if you're using a library, in your application, that is instrumented with metrics, you get that metric data "for free". It could be useful to include the metrics emitted by dependencies as part of an exemplar, but you can't do that if they need to use a specific macro.

Alright, my thinking behide the current poc implementation and suggestions was to keep all things that are specific to prometheus in the prometheus-exporter, but you have a good point that you then wouldn't get those specific features out of the box when using libraries that exposes metrics via this lib.

> What I would want to see is figuring out a way to make the determination of when exemplars should be tracked something that functions more like a scoped behavior i.e. a function that takes a closure, and changes a thread-local to influence the behavior of any code running the closure

I'm not sure how to implement that, so I'll have to do some digging, if there is any such implementations in this or other crates that you know about, I would appreciate any pointers.

> I'm happy to continue providing feedback on the design as long as it's along the lines as described above, because that represents both the most ergonomic path, and easiest to maintain path, in my eyes. If executing on what I've posted above feels like it would consume too much time, I totally understand.

Thanks, I'll keep at it whenever I have time to spare, this would be a really useful feature for trace discoverability.

bowman2030:
@tobz 
> it doesn't feel right for there to be macros/methods/etc that are specific to Prometheus

I think there's a misunderstanding here and I'd really like the chance to add context to the discussion. Please take all the following as constructive input from a would-be fan. To some users like me, this would be the difference between always using and eagerly recommending `metrics`, or going out of my way to warn people to avoid it as a known dead end. I know that might sound like an over-reaction, but I hope it'll be clear where I'm coming from by the end of this post.


First to fill in some gaps in the discussion:

* [OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md) is bigger than Prometheus, it's preserving what worked with Prometheus but expanding it to more features and more ingestion implementations. It seems to have [stalled](https://datatracker.ietf.org/doc/draft-richih-opsawg-openmetrics/) as an IETF standard, but it's already implemented in a number of exposition and ingestion systems, so it's at the very least a de facto standard. It is safe to consider this the new lowest common denominator for metric systems, and that lowest common denominator includes exemplars.

* The fact it's not a ratified standard is a small mark against it, but it's still much closer to a standard than the pre-OpenMetrics Prometheus exposition format that you already support and that `metrics` is functionally just a reasonable subset of. The Prometheus format's lack of specification or standardization didn't stop you or anyone else from supporting it, and OpenMetrics goes a lot further in specifying everything you need to know for exposition, ingestion, and applied usage, so if anything it's a firmer foundation on which to build a united front like this facade.

* Exemplars are the compromise between timeseries databases with many samples of small label sets, and event databases with fewer samples but larger label sets. Exemplars let individual samples have more labels without blowing out the dimensionality of the timeseries as a whole. It is not specific to OpenMetrics, much less to Promethus. It's a reality that every timeseries database has to face, otherwise it'll be abandoned in favor of a more powerful event database like InfluxDB.

* OpenTelemetry is even more ambitious than OpenMetrics, and also [supports exemplars](https://opentelemetry.io/docs/reference/specification/metrics/data-model/). It's also just a specification, and maybe one day it will be formalized as a standard, but the industry is rallying around it already.

* So both relevant specifications in this space support exemplars. Other specifications and implementations can reasonably expect exemplars in the data. Whatever standards come out of all of this, going forward, I can't imagine they'll leave exemplars out ever again. It would be like leaving out histograms.

* Of course, not every implementation has caught up, but that's absolutely on them. There are browsers out there that don't support JS, and we just don't use those browsers, their existence doesn't somehow mean JS itself never needs to be written or executed again. At one point in the past there was only one browser that supported JS, now there are specifications and multiple implementations, it wouldn't be accurate or useful to dismiss JS as "specific to" any of them.

* The [prometheus_client](https://github.com/prometheus/client_rust) crate already supports exemplars today. (It was [named](https://github.com/tikv/rust-prometheus/issues/392#issuecomment-968691129) for consistency with other libraries, but it is clearly [billed as an OpenMetrics exposition library](https://github.com/prometheus/client_rust#prometheus-rust-client-library)).

* To a Rust developer, a metrics facade may be the point of interoperability everything hinges around. But zooming out, across languages and libraries and so on, OpenMetrics and OpenTelemetry are those points of interoperability, little else is relevant on the same scale. If data needs to be exposed, code will be written to expose it, and if `metrics` makes that impossible then that code just won't use `metrics`. Implementation details are negotiable, functional requirements often aren't.

* Already, today, a project can use `prometheus_client` to ensure all useful information is available for monitoring. _At worst_ they can translate from OpenMetrics to another backend, perhaps using a standalone service that isn't even specific to Rust. Writing that translation once wouldn't be much more work than just writing a recorder just for `metrics` specifically. Some deployments I know already use Telegraf for such a translation today.

* Thus OpenMetrics has the most leverage in shaping ecosystem choices, libraries at best get to save people some trouble in exposing OpenMetrics, and `prometheus_client` offers more here than `prometheus` or `metrics` at this point in time.


Look, if someone's hobby project supported parsing base64 reaction meme gifs from metrics to render inline in alerts, I would understand `metrics` not supporting that because it's a niche, ad-hoc, unspecified, likely unheardof extension. But exemplars are not that, they are a part of multiple published specifications each with multiple mainstream implementations enjoying wide industry adoption.

You've mentioned in several places that you wanted `metrics` to follow `log` and `tracing`, which sounds good on paper, but I think the fact there are two should serve as a warning. If `metrics` does not grow to support the features people need in this space, then at best there'll be another facade that does, just like `tracing` does strictly more than `log`. However, that would be a strictly worse situation, because it will clutter the ecosystem indefinitely as all the existing macros and implementations would still have to be supported in addition to the new ones. The whole ecosystem is better off having just one `metrics` that does everything and we can all rally around.

It's hard enough explaining to folks why both `log` and `tracing` exist, and I know lots of people are intimidated by the scope of `tracing` even though it's no harder to use than `log` for most purposes. Rust already gets this reputation for all sorts of things, monitoring shouldn't be another one. I want it to be easy for everyone to export metrics, but I don't want anyone investing in a dead-end API that they'll have to migrate off as soon as their needs outgrow what `metrics` is willing to support.

**Looking forward**

Now let me offer constructive suggestions on how to go about this. I think this PR would have gone very differently if `metrics-exporter-prometheus` used the new `prometheus_client` instead of the known-dead-end `prometheus` crate. Then the underlying support for exemplars would have already been there, and it would only be a matter of providing an API around that.

I don't see a way around migrating from `prometheus` to `prometheus_client`. Proposals to support OpenMetrics in general and exemplars in particular for `prometheus` were shot down specifically because `prometheus_client` already solved these problems in code that ships today. Given `prometheus` is a dead end, `metrics` will have to have a `prometheus_client` recorder sooner or later. When that inevitably happens, it will be much easier to support exemplars, and much harder to rationalize refusing to support them. In other words, `metrics` has the choice of being left behind with `prometheus` or to evolve alongside `prometheus_client`.

Maybe `metrics v1.0.0` can launch without exemplars then e.g. `v1.1.0` can add them in a backwards compatible way. The Go library showed just how wrong this can go, but this has more chance of succeeding in Rust than Go because at least Rust traits can evolve using default method implementations (i.e. recorders can ignore exemplars in the default implementation and add support if and when possible).

Personally I still think `v1.0.0` is the perfect opportunity for the ecosystem to rally around a facade that does support exemplars, and what better time to introduce the world to a recorder based on `prometheus_client` which will remain the right way forward to capitalize on OpenMetrics without leaving vital functionality out in the cold. I don't think anybody will care if it took a few more months to get to that point, I think way more people will care whether it successfully captured the future of the ecosystem.

Thank you for your patience with this wall of text. I spent hours today writing this because, if there's any chance this sends `metrics` in a better direction, it will have been well worth it.

Issue #323, opened 2022-08-17
Having a default of std-atomics is problematic for 32 bit archs


huntc:
Further to https://github.com/metrics-rs/metrics/pull/313, I believe we have a problem given that std-atomics are a default feature for the metrics crate. As with any feature, you cannot "unset" it. Therefore, when depending on, say, metrics-utils, the default feature will be enabled no matter what your target is.

The std-atomics default was added back in for https://github.com/metrics-rs/metrics/commit/65d965c10591373cf7348570c45c1d156fb5f92c.

Perhaps I'm missing something though? Or would it be better to have the std-atomics feature as the default for the other "leaf" crates?

That said, shouldn't the portable atomic library enable Atomic64 where it is available i.e. do we need the std-atomics feature at all?

For completeness, I'm trying to depend on both metrics-exporter-prometheus and metrics-util.

huntc:
> So, at a high level: I don't intend to force everyone using `metrics` to use `portable-atomic`. Practically speaking, does it provide equivalent behavior to `std` for architectures that don't need shimmed atomics? Yes. Do I want to force that dependency on them, though? No.
> 
Isn’t portable atomic already mandatory as a dependency?

Also, I think we should either trust portable atomic or not. I’m happy with not, but we will then need to reconsider the use of 64 bit operations and/or perhaps the applicability of metrics for 32 bit archs. Wdyt?

tobz:
I do trust `portable-atomic`, but forcing every implementor to use it just... doesn't make sense. There's no technical reason why the atomics from the standard library aren't good enough, if that's what someone prefers.

I do realize that this perhaps feels like circular logic/silliness -- "you told me `portable-atomic` was good enough, and now it's not?" -- but I do believe this is the best middleground for now, so long as we make `metrics` leaf crates opt out of the default features: everyone gets the same available impls for `CounterFn`, etc, as they've had since the traits were introduced, but `metrics` leaf crates can scope themselves down so they always work seamlessly on 32- or 64-bit targets.... and external crates can make whatever decision they choose re: what types to use, what archs to support, etc.

huntc:
> I do trust `portable-atomic`, but forcing every implementor to use it just... doesn't make sense. There's no technical reason why the atomics from the standard library aren't good enough, if that's what someone prefers.

Sure, but aren’t we getting the std lib atomics still with portable atomic when supported by the target? I believe we are but my understanding could be off. 

> 
[...]

PR #324, opened 2022-08-17
Removes the std-atomic feature

huntc:
The std-atomic feature is not required as the portable atomics library will fall back to native atomics when there are available.

Additionally, having a std-atomics feature as a default prevents the other metrics libraries such as metrics-util, from being used on 32 bit archs.

Fixes https://github.com/metrics-rs/metrics/issues/323

### DIFF
--- a/metrics/Cargo.toml
+++ b/metrics/Cargo.toml
@@ -25,10 +25,6 @@ bench = false
 name = "macros"
 harness = false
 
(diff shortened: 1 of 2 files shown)


tobz:
This should be solved via a3a706b.

huntc:
Thanks for solving this issue!

PR #345, opened 2023-02-27
Bring the metrics-observer protobufs up to date

### LINKED ISSUE #325: metrics-observer is not available
run metrics-observer have DecodeError.

I found that  [metrics-observer/proto/event.proto](https://github.com/metrics-rs/metrics/blob/64d5c5ae63babafbf3875f723ed263efe9332647/metrics-observer/proto/event.proto) and [metrics-exporter-tcp/proto/event.proto](https://github.com/metrics-rs/metrics/blob/64d5c5ae63babafbf3875f723ed263efe9332647/metrics-exporter-tcp/proto/event.proto) are inconsistent

ceejbot:
Aligned the proto file in the observer sub-crate with the one in the TCP exporter. Updated the observer message handling code to use the new structures. 

This format de-sync will occur again if the writer code is changed without also updating the observer. The right solution here is to have the two crates share the protobuf file, but that requires more structural tinkering than a newcomer like me should do :). 

Satisfied the clippy lints complaining about type complexity by creeating some subtypes.

I note that in order to test with rustc 1.56.1, I had to pin the version of `mockall` used by metrics-util to `0.11.0`.

This fixes #325 "metrics-observer is not available".


metrics-observer/src/metrics.rs
```diff
@@ -35,10 +32,14 @@ pub enum MetricData {
     Histogram(Summary),
 }
 
+// Intermediate types used to satisfy clippy's dislike of complexity.
+type MetadataKey = (MetricKind, String);
+type MetdataValue =  (Option<Unit>, Option<String>);
```
tobz, line 37: `MetdataValue` -> `MetadataValue`

metrics-observer/src/metrics.rs
```diff
@@ -16,11 +16,8 @@ mod proto {
     include!(concat!(env!("OUT_DIR"), "/event.proto.rs"));
 }
 
-use self::proto::{
-    event::Event,
-    metadata::{Description as DescriptionMetadata, MetricType, Unit as UnitMetadata},
-    Event as EventWrapper,
-};
+use self::proto::metadata::MetricType;
+use proto::event::Event as ProtoEvent;
```
tobz, line 20: Not clear why we're importing `Event` as an aliased name here? There's no conflict I can see.

tobz (changes requested):
I've left a few comments about specific changes, but generally, I'd like to see qualified references (i.e. `match proto::event::Event::decode { ...`) be changed to drop the qualified portion (`proto::event`) and move those types up to the imports at the top of the file.

This keeps things cleaner in the business logic and is the prevailing style throughout the project.

Generally speaking, though, these changes make sense and it'll be good to get the TCP exporter back on track.

metrics-observer/src/metrics.rs
ceejbot: There is a lurking conflict between the protobufs message wrapper and the metrics event type, but I think the latest commit clarifies this with a better name.

tobz:
Released as `metrics-exporter-tcp@v0.8.0` and `metrics-observer@v0.2.0`.

PR #347, opened 2023-03-11
Use `portable-atomic` only for 32-bit architectures

### LINKED ISSUE #323: Having a default of std-atomics is problematic for 32 bit archs

### LINKED ISSUE #346: Just use only std atomic without `portable-atomics`?
Seems even when using the `std-atomics` feature the `portable-atomics` crate is still included and required as it is part of the public interface? And not re-exported in the crate either so users `metrics` with 64-bit counters have to manually import that crate with the matching version (0.3, latest is 1.0 which is incompatible). 

This feels somewhat backwards, would it be possible that purely std atomics are used by default without bringing in `portable-atomics` (as in reverse the feature flag)?

cc @marcusgrass

repi:
A new take on how to use the atomics without bringing in the `portable-atomic` crate for 64-bit architectures where one can use just `std`.  

Believe this approach could satisfy the requirements, and be easier to use for everyone, but I'm not super familiar with this crate so may have missed something also.

Resolves #346
Resolves #323

### Changed

- Only use `portable_atomic::AtomicU64` for 32-bit platforms, for 64-bit platforms use std.

### Added

- Re-export `AtomicU64` so one have an easy and convenient way to get the used atomic counter type regardless if one 32-bit or 64-bit architectures and without depending on external crates when using this crate.

### Removed

- Removed the `std-atomics` Cargo feature as it is not longer needed

### DIFF
--- a/metrics-benchmark/Cargo.toml
+++ b/metrics-benchmark/Cargo.toml
@@ -12,6 +12,5 @@ pretty_env_logger = "0.4"
 getopts = "0.2"
 hdrhistogram = { version = "7.2", default-features = false }
 quanta = "0.10.0"
-portable-atomic = "0.3"
 metrics = { version = "^0.20", path = "../metrics" }
 metrics-util = { version = "^0.14", path = "../metrics-util" }
--- a/metrics-benchmark/src/main.rs
+++ b/metrics-benchmark/src/main.rs
@@ -3,10 +3,9 @@ use hdrhistogram::Histogram as HdrHistogram;
 use log::{error, info};
 use metrics::{
     gauge, histogram, increment_counter, register_counter, register_gauge, register_histogram,
-    Counter, Gauge, Histogram, Key, KeyName, Recorder, SharedString, Unit,
+    AtomicU64, Counter, Gauge, Histogram, Key, KeyName, Recorder, SharedString, Unit,
 };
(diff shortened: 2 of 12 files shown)


tobz:
I ended up replicating this by hand in a3a706b as I wanted to slightly tweak some things. I'm going to close this PR with that in mind, but I did make sure to credit you for that change in the CHANGELOG, as it ultimately overlapped with your diff by nearly 100%.

Thanks again for your contribution and kicking my butt to get this solved in a cleaner way.

repi:
thanks!

tobz:
Released as `metrics@v0.21.0`.

PR #348, opened 2023-03-13
Add white background to splash image

Xaeroxe:
I'm going to make an assumption and say that this presentation in GitHub dark mode is not what you intended, and add a white background to your splash image.
Screenshot 2023-03-13 at 12 14 12 PM


tobz:
It is indeed not what I intended, although big ass white blob in dark mode is also stylistically not great 🙈 ... although perhaps better, all things considered, than the current situation. 😂

tobz:
Thanks for your contribution. ❤️

PR #359, opened 2023-04-17
chore: add global preifix for recorder

paomian:


tobz:
This should already be possible utilizing the [prefix layer](https://docs.rs/metrics-util/latest/metrics_util/layers/struct.Prefix.html) from `metrics-util`.

I realize that the Prometheus exporter already supports setting global labels, but the exporter is already at the limit of customization that I'm comfortable with, and I want to encourage more usage of the existing utilities (such as layers) to avoid exporter feature fragmentation.

Happy to re-open the PR if you believe there's a compelling reason why this should exist at the exporter level.

Issue #308, opened 2022-06-27
How to avoid data loss on shutdown


jorendorff:
:wave: Hi, everyone. I work on code search at GitHub and we ❤️ `metrics`.

We have a panic hook that does `metrics::increment_counter!("panic");`. But we've found that this increment often doesn't make it to the server.

I think our recorder (we're using [`metrics-expoter-statsd`](https://github.com/github/metrics-exporter-statsd/)) buffers output and the process is exiting without flushing the buffers. Dropping the recorder before exiting the process would be sufficient, for us. Is there a way to do that? Is there something else I should be doing? I couldn't figure it out from the readme or API docs.

If this already exists, I think there's already a great place to document it (under "Getting started -> In executables") and I'd be happy to write that up.

tobz:
Hello there!

Reading through your description, I see where you're coming from.

At a high level, there is no way, currently, to drop the recorder as it's installed in such a way that callers can take a `'static` reference to it, which means that installing a recorder requires either a `&'static dyn Recorder` or an owned `R: Recorder` which we box and then leak, in order to fulfill the `'static` requirement.

Most of the ways around this that I can think of would depend on the exporter being able to give back a handle, of some sort, to the caller that is creating/installing the exporter such that the handle would act as the drop guard that could be used to force a flush, etc, when the process is exiting.

tobz:
Another thought is potentially doing something with `Arc<T>`/`Weak<T>` where the weak reference can be wrapped and installed as the recorder and then then `Arc<T>` could be held as a drop guard so that the true `R` (being `R: Recorder`) could be legitimately dropped and have the `Drop` impl added directly, and not need to write a custom drop guard, with all of the requisite synchronization.

This would still necessitate holding that drop guard directly, rather than the installed recorder being able to handle this, but it would at least be something we could provide a generalized wrapper for.

tobz:
@jorendorff A bit late, but... I have a PR up for a drop guard primitive, as described in my last comment on this issue, that I'm curious if you would be interested in. At least ,if nothing else, I'm curious to see if this _would_ have helped solve your problem (if you already found a solution and don't care to switch away from it).

aig787:
Hey @tobz, not OP, but I was facing this same issue with an influx/grafana cloud exporter and fixed it using this approach. Any plans to cut a new release?

tobz:
Yeah, I'm a little behind on releasing the currently outstanding changes. Hoping to be able to do so this week.

[...]

PR #362, opened 2023-04-18
feat(util): new helper type for recovering recorder after installing it

tobz:
## Context

In cases like #308, there can be a desire to ensure that a recorder flushes itself on process exit. As we install the recorder globally, such that it is leaked to provide the ability to acquire a static reference to it, recorders cannot normally participate in the typical `Drop` logic as objects are cleaned up before process exit.

## Solution

This PR introduces a generic helper -- `RecoverableRecorder<R>` -- that allows for installing a wrapped version of the recorder, and acts as a drop guard to allow either manually "recovering" the recorder, or recovering it on drop such that the recorder's drop logic can also be called.

We wrap the recorder in an `Arc<T>`, and install a custom recorder wrapper that holds a `Weak<T>` reference. This allows the global recorder to provide access to the inner recorder without taking ownership of it.

When the caller manually seeks to recover the recorder, they call `RecoverableRecorder::into_inner` which attempts to unwrap the `Arc<T>` until there are no active references other than from the drop guard, which gives back the original recorder. If the drop guard is simply dropped, the original recorder too is dropped as soon as all active references to it have dropped, following the normal behavior of `Arc<T>`.

This does come with a small performance overhead, as every call through the weakly-held `Recorder` implementation must clone the `Arc<T>` before proxying the given method call, but it should be fairly minimal in practice. If performance is a huge concern, a recorder should implement its own drop guard for controlling flushing on process exit, but this is simply meant to be a generic solution when the application has no control over the recorder implementation.

Additionally, I've stripped out all of the `atomic_cas` and `set_recorder_racy` bits, as all illusions of `metrics` being `no_std`/embedded-compatible have sort of faded at this point, and it was just an opportune time to strip out the last vestiges.


tobz:
Released in `metrics-util@v0.15.1`.

Issue #365, opened 2023-04-27
push-gateway does not support passing authentication information


JasonLi-cn:
https://github.com/metrics-rs/metrics/blob/8e2e39df52c911563c7f38937b31f2f4c6e28cba/metrics-exporter-prometheus/src/builder.rs#L65

Currently, pus-gateway only supports passing uri and interval, but not username and password.

tobz:
Addressed via #366.

PR #366, opened 2023-04-27
push-gateway support authentication

JasonLi-cn:
#365 

In some cases where push-gateway is used, we need to pass in username and password.

### DIFF
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -28,6 +28,7 @@ metrics-util = { version = "^0.15", path = "../metrics-util", default-features =
 thiserror = { version = "1", default-features = false }
 quanta = { version = "0.11", default-features = false }
 indexmap = { version = "1", default-features = false }
+base64 = { version = "0.21.0" }
 
 # Optional
 hyper = { version = "0.14", default-features = false, features = ["tcp", "http1"], optional = true }
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -28,7 +28,7 @@ metrics-util = { version = "^0.15", path = "../metrics-util", default-features =
 thiserror = { version = "1", default-features = false }
 quanta = { version = "0.11", default-features = false }
 indexmap = { version = "1", default-features = false }
-base64 = { version = "0.21.0" }
+base64 = { version = "0.21.0", default-features = false, features = ["std"] }
 
 # Optional
 hyper = { version = "0.14", default-features = false, features = ["tcp", "http1"], optional = true }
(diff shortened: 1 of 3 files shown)


metrics-exporter-prometheus/src/builder.rs
```diff
@@ -531,6 +547,28 @@ impl Default for PrometheusBuilder {
...
+        let _ = write!(encoder, "{username}:");
+        if let Some(password) = password {
+            let _ = write!(encoder, "{password}");
+        }
+    }
+    let mut header = HeaderValue::from_bytes(&buf).expect("base64 is always valid HeaderValue");
+    header.set_sensitive(true);
+    header
+}
+
```
tobz, line 571: This should have a simple unit test below that ensures the output matches well-known examples for both username only, and username/password.

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -531,6 +547,28 @@ impl Default for PrometheusBuilder {
     }
 }
 
+fn basic_auth<U, P>(username: U, password: Option<P>) -> HeaderValue
```
tobz, line 550: Just make these `&str` and `Option<&str>`. No reason to be generic.

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -22,6 +22,7 @@ use hyper::{
     Response, StatusCode,
 };
 
+use hyper::http::HeaderValue;
```
tobz, line 25: Should add this to the `cfg`-gated import below, since it's only used when push gateway support is enabled.

metrics-exporter-prometheus/src/builder.rs
tobz: And also that if the return value (`HeaderValue`) is rendered via its debug representation (`{:?}`), the value is `Sensitive`.

(I realize it's very obvious from the code that this should happen, but since we're not using `typed-headers`, I want to make sure we're validating basic properties via tests)

metrics-exporter-prometheus/Cargo.toml
```diff
@@ -28,6 +28,7 @@ metrics-util = { version = "^0.15", path = "../metrics-util", default-features =
 thiserror = { version = "1", default-features = false }
 quanta = { version = "0.11", default-features = false }
 indexmap = { version = "1", default-features = false }
+base64 = { version = "0.21.0" }
```
tobz, line 31:
```suggestion
base64 = { version = "0.21.0", default-features = false, features = ["std"] }
```

tobz (changes requested):

JasonLi-cn:
@tobz please review

tobz:
The feature check stuff is going to be tedious, so I'll merge for now and sort that out locally.

Issue #370, opened 2023-05-12
Consider adding richer metadata to metric emission.


tobz:
In #338, it was mentioned/implied that metric filtering is harder than it needs to be due to a lack of first-class metadata, such as what [`tracing`](https://docs.rs/tracing) provides in its "[callsite metadata](https://docs.rs/tracing/latest/tracing/struct.Metadata.html)". This includes high-level things like log level, but also target, module path, and so on.

These additional metadata properties allow users to efficiently filter out, for example, all spans/events from a specific crate or module. In contrast, with `metrics`, at best, we can filter by metric name and/or labels, which is precarious and involves a priori knowledge of exactly what we want to filter.. whereas tracing's approach can allow (at least far more easily) for filtering via allowlist rather than blocklist.

We should consider what a design might look like to bring this same callsite metadata to `metrics`.

The primary concerns would roughly be:

- what data should be included? (crate/module path seem obvious, but would "level" make sense?)
- should this data be passed alongside `Key` or baked into it?

Issue #337, opened 2022-11-16
Provide a way to set unit without setting description


jaskij:
As far as I can tell, currently the only way to set a unit is to use the `describe_histogram!()` macro. This macro requires three arguments (third being description) and if you try to call it with only two, as `describe_histogram!(Self::HISTO_NAME, metrics::Unit::Seconds);`, it errors out with an extremely cryptic error about a missing semicolon.

A temporary workaround is to simply set an empty description, but the error over what, to me, is an intuitive use of the API, is extremely confusing.

tobz:
You're right that the error is cryptic, which is a side effect of using procedural macros, in this case.

The main problem is that the actual `Recorder` trait doesn't allow for the description to be optionally set. The trait could likely be changed to make the description optional, along with the `describe_*` macros to take both unit or description separately and optionally. However, my time is a bit fleeting these days so I can't commit to doing so in the near-term.

PRs are definitely welcome and I'd be happy to guide you, or anyone, working on such a PR.

tobz:
Mulling this over, I'm going to see if maybe there's a more generic approach we can take in the macros. My thought is that "describing" a metric could/should be more like applying labels to it, such that a hypothetical refactor of `describe_counter!` could look something like this:

`describe_counter!(description => "blah blah", unit => Unit::Nanoseconds)`

Ultimately, certain recorders (or certain applications) might care about descriptions, some might only care about units, or they might care about entirely different things altogether. The current design is limiting in that regard, notwithstanding the issue you've brought up here around having to specify a description just to set a unit.

I want to think about whether or not there's a way to, perhaps, make it less of a raw key/value map of attributes where every recorder needs to, for example, hope/assume the describe calls use `description` for the description attributes and so on. Maybe there could be some sort of enum/fixed constants to standardize around for declaring common attributes, and then a catch-all variant for describing "custom" attributes... not sure.

Anyways, yeah, I'm going to give this some thought.

tobz:
Yeah, I'm familiar with `valuable`, and overall, your sketch is what I was roughly envisioning when typing up my previous reply.

I'm still torn on what should be "standard", and whether or not `metrics` should enforce it. As an example, is `metrics::Unit` actually sufficient for all users as-is, or should it also get the `valuable` treatment?

[...]

PR #380, opened 2023-06-06
Add metadata to metrics

hlbarber:
**Motivation**

Closes https://github.com/metrics-rs/metrics/issues/370

**Changes**

- Add `&Metadata` argument to the `Recorder::register_*` methods
- Add `target: *` and `Level::*` arguments to macros.
    - `target` defaults to `module_path!()`
    - `Level` defaults to `Level::INFO`.

**TODO**

- [ ] Add documentation
- [ ] Add `info_counter!`, `warn_counter!`, etc
- [ ] Improve test coverage

### DIFF
--- a/metrics-exporter-prometheus/src/builder.rs
+++ b/metrics-exporter-prometheus/src/builder.rs
@@ -579,13 +579,21 @@ mod tests {
 
     use super::{Matcher, PrometheusBuilder};
 
+    static METADATA: metrics::Metadata = metrics::Metadata::new(
+        module_path!(),
+        metrics::Level::INFO,
+        Some(module_path!()),
+        Some(file!()),
+        Some(line!()),
+    );
+
     #[test]
     fn test_render() {
         let recorder =
             PrometheusBuilder::new().set_quantiles(&[0.0, 1.0]).unwrap().build_recorder();
 
         let key = Key::from_name("basic_counter");
-        let counter1 = recorder.register_counter(&key);
+        let counter1 = recorder.register_counter(&key, &METADATA);
         counter1.increment(42);
 
         let handle = recorder.handle();
@@ -596,7 +604,7 @@ mod tests {
 
         let labels = vec![Label::new("wutang", "forever")];
         let key = Key::from_parts("basic_gauge", labels);
-        let gauge1 = recorder.register_gauge(&key);
+        let gauge1 = recorder.register_gauge(&key, &METADATA);
         gauge1.set(-3.14);
         let rendered = handle.render();
         let expected_gauge = format!(
@@ -607,7 +615,7 @@ mod tests {
         assert_eq!(rendered, expected_gauge);
 
         let key = Key::from_name("basic_histogram");
-        let histogram1 = recorder.register_histogram(&key);
+        let histogram1 = recorder.register_histogram(&key, &METADATA);
         histogram1.record(12.0);
         let rendered = handle.render();
 
@@ -649,19 +657,19 @@ mod tests {
             .build_recorder();
 
         let full_key = Key::from_name("metrics.testing_foo");
-        let full_key_histo = recorder.register_histogram(&full_key);
+        let full_key_histo = recorder.register_histogram(&full_key, &METADATA);
         full_key_histo.record(FULL_VALUES[0]);
 
         let prefix_key = Key::from_name("metrics.testing_bar");
-        let prefix_key_histo = recorder.register_histogram(&prefix_key);
+        let prefix_key_histo = recorder.register_histogram(&prefix_key, &METADATA);
         prefix_key_histo.record(PREFIX_VALUES[1]);
 
         let suffix_key = Key::from_name("metrics_testin_foo");
-        let suffix_key_histo = recorder.register_histogram(&suffix_key);
+        let suffix_key_histo = recorder.register_histogram(&suffix_key, &METADATA);
         suffix_key_histo.record(SUFFIX_VALUES[2]);
 
         let default_key = Key::from_name("metrics.wee");
-        let default_key_histo = recorder.register_histogram(&default_key);
+        let default_key_histo = recorder.register_histogram(&default_key, &METADATA);
         default_key_histo.record(DEFAULT_VALUES[2] + 1.0);
 
         let full_data = concat!(
@@ -724,15 +732,15 @@ mod tests {
             .build_with_clock(clock);
 
         let key = Key::from_name("basic_counter");
-        let counter1 = recorder.register_counter(&key);
+        let counter1 = recorder.register_counter(&key, &METADATA);
         counter1.increment(42);
 
         let key = Key::from_name("basic_gauge");
-        let gauge1 = recorder.register_gauge(&key);
+        let gauge1 = recorder.register_gauge(&key, &METADATA);
         gauge1.set(-3.14);
 
         let key = Key::from_name("basic_histogram");
-        let histo1 = recorder.register_histogram(&key);
+        let histo1 = recorder.register_histogram(&key, &METADATA);
         histo1.record(1.0);
 
         let handle = recorder.handle();
@@ -774,15 +782,15 @@ mod tests {
             .build_with_clock(clock);
 
         let key = Key::from_name("basic_counter");
-        let counter1 = recorder.register_counter(&key);
+        let counter1 = recorder.register_counter(&key, &METADATA);
         counter1.increment(42);
 
         let key = Key::from_name("basic_gauge");
-        let gauge1 = recorder.register_gauge(&key);
+        let gauge1 = recorder.register_gauge(&key, &METADATA);
         gauge1.set(-3.14);
 
         let key = Key::from_name("basic_histogram");
-        let histo1 = recorder.register_histogram(&key);
+        let histo1 = recorder.register_histogram(&key, &METADATA);
         histo1.record(1.0);
 
         let handle = recorder.handle();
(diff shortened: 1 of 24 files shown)


hlbarber:
Added this in a draft state to garner comments before polishing it up.

tobz:
@hlbarber High-level thoughts: this is pretty straightforward, but the macros should probably look more like the `tracing` macros.

In and of itself, `Metadata` feels as straightforward as I imagined it to be, but the lack of macro examples and the copied approach of how the `describe_*` macros handle detecting the unit parameter have solidified my thoughts around the macros needing to handle this better and in a way that's entirely compatible with the current macros. This is separate from actually providing level-prefixed variants of the macros, which I'm still a little hesitant about.

Essentially, what I'd want to see is:

- `target` and `level` require a struct field notation in the macro arguments (like `tracing`)
- metric name is always the first argument
- parse the remaining arguments as either the "fixed function" `target` or `level` fields, or as labels (label parsing logic stays as-is)
- cannot specify `target` or `level` _after_ we've parsed something label-like (set a flag once we see the first non-`target`/`level` arg, basically)

Or with examples:

```
register_counter!("metric_name");
register_counter!("metric_name", "label_a" => "...");
register_counter!("metric_name", target: "overridden");
register_counter!("metric_name", level: Level::TRACE, &some_vec_of_labels);
register_counter!("metric_name", target: "overridden", level: Level::Info, "service" => "foo");
```

Ultimately, seeing this PR, I'd say that you could keep going with everything else on your list but hold off on the `info_counter!` stuff. If you're willing to take a crack at the procedural macro changes I've proposed above, that'd be good to. Otherwise, I would likely submit those as changes directly to your PR in the next week or two.

hlbarber:
[`tracing::span`](https://docs.rs/tracing/latest/tracing/macro.span.html) and [`tracing::event`](https://docs.rs/tracing/latest/tracing/macro.event.html), in order, accepts:

- Optional target: `target: {target}`
- Required level: `Level::{LEVEL}`
- Required name: `"{name}"` (for `span!`)

This implementation, in order, accepts:

- Optional target: `target: {target}`
- Optional level: `Level::{LEVEL}` (for backwards compatibility)
- Required metrics name: `"{name}"`
- ...

```rust
register_counter!("metric_name");
register_counter!("metric_name", "label_a" => "...");
register_counter!(target: "overridden", "metric_name");
register_counter!(Level::TRACE, "metric_name", &some_vec_of_labels);
register_counter!(target: "overridden", Level::Info, "metric_name", "service" => "foo");
```

> target and level require a struct field notation in the macro arguments

Agree, I can change `level: Level::{LEVEL}` _or_ make `Level::{LEVEL}` required (like `tracing`) - either of these will avoid the "unit parameter approach" (which I agree should perhaps be avoided). I think the struct field notation helps for optional fields.

_Side note: Perhaps the unit parameter should be struct field notation too? Maybe this allows these macros to be done declaratively rather than procedurally._

> metric name is always the first argument

Can do, but this seems like a departure `tracing` where

```rust
macro_rules! span {
    (target: $target:expr, parent: $parent:expr, $lvl:expr, $name:expr) => { ... };
    (target: $target:expr, parent: $parent:expr, $lvl:expr, $name:expr, $($fields:tt)*) => { ... };
```

tobz:
> I can change `level: Level::{LEVEL}` _or_ make `Level::{LEVEL}` required (like `tracing`) - either of these will avoid the "unit parameter approach" (which I agree should perhaps be avoided). I think the struct field notation helps for optional fields.

Right, I want `level` to use the struct field notation because it it's optional.

> _Side note: Perhaps the unit parameter should be struct field notation too? Maybe this allows these macros to be done declaratively rather than procedurally._

The hard-coded bits of the `describe_*!` macros are going away with the rework to metric "attributes" -- https://github.com/metrics-rs/metrics/issues/337#issuecomment-1511547332 -- so I would just ignore that for the moment.

> Can do, but this seems like a departure `tracing` where

Fair point. I wasn't thinking we'd necessarily mirror everything about how the `tracing` macros do it, but I suppose all named optionals having to come first would make parsing easier, especially if we attempted to rewrite the macros to be declarative.

I'm also not against doing that either -- proc macros are heavy -- but it might need to be staged as an improvement that comes after this and metrics attributes, because those two items get rid of the grossest parts of the procedural macros, which is the wishy-washy handling is specific positional arguments (i.e. is this arg a `Unit`? let's check some hardcoded paths).

metrics/src/metadata.rs
```diff
@@ -0,0 +1,46 @@
...
+    Info = 2,
+    Warn = 3,
+    Error = 4,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Metadata<'a> {
+    target: &'a str,
+    level: Level,
+    module_path: Option<&'a str>,
```
tobz, line 27: Hmm, why exactly is module path optional but not target?

Seems like we could swap that around, and then have `target()` return `self.target.as_ref().unwrap_or(self.module_path)` or something like that.

metrics/src/metadata.rs
```diff
@@ -0,0 +1,46 @@
+#![allow(missing_docs)]
```
tobz, line 1: Obviously this was in place for the draft aspect, but just noting it that we'll need to drop this and add docs for all the new public types here.

tobz (changes requested):
After sitting down and going over this, I feel a lot better about the nature of the change and what we're adding.

Left two small comments.

metrics/src/metadata.rs
hlbarber: I've just imitated `tracing`s behavior here.

https://github.com/tokio-rs/tracing/blob/e603c2a254d157a25a7a1fbfd4da46ad7e05f555/tracing-core/src/metadata.rs#L68-L70

> The name of the Rust module where the span occurred, or `None` if this could not be determined.

Admittedly I don't know _when_ this cannot be determined given that `module_path!` returns a `&'static str`.

I will do some investigation about the reason why `tracing::Metadata` is setup like this and then get back to you.

hlbarber: Summarizing my brief discussion with the `tracing` maintainers:

- `module_path` should _always_ be populated when the `Metadata` is constructed from the `tracing` macros.
- It is represented as an `Option`:
    1. To maintain compatibility with `log` whose [`module_path`](https://docs.rs/log/latest/log/struct.Record.html#method.module_path) is also optional.
    1. To interop with events emitted in contexts where `module_path` might not be readily accessible (outside of Rust).

We don't need to be compatible with `log` however it might be nice to enable the use case (ii).

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -579,13 +579,16 @@ mod tests {
 
     use super::{Matcher, PrometheusBuilder};
 
+    static METADATA: metrics::Metadata =
+        metrics::Metadata::new(module_path!(), metrics::Level::INFO, Some(module_path!()));
```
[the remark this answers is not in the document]
tobz: Eh, it's hard to see the value at this point, given that there's only three parameters needed. I guess, maybe, if we added something like `metrics::default_metadata!()`?

Avoiding needing to type out the `module_path!()` macro in the right place would be one thing, but I don't know if the complexity would be worth it to make it support a builder pattern (i.e. use the default level but set a custom target) over just having people call `Metadata::new` themselves... or providing builder pattern methods on `Metadata`.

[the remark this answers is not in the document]
hlbarber: > providing builder pattern methods on Metadata

Regarding this, would you prefer it? I've been quite conservative with what we've added to `Metadata`. The `Metadata::new` obviously ties us to the current set of fields, if you anticipate perhaps adding something like [`Metadata::file`](https://docs.rs/tracing/latest/tracing/struct.Metadata.html#method.file) in the future then `Metadata::builder` is probably preferable?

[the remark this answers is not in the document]
tobz: Yeah, let's just wait.

While the builder methods would make it easier to write a declarative macro (since we could just collect each build method fragment and smoosh them together at the end) there's still a little bit tied up in setting/specifying the defaults for the current field and it's not clear to me what the sharp edges, if any, will be with the current approach... so I think we probably just need to use it first.

metrics/src/metadata.rs
```diff
@@ -0,0 +1,65 @@
...
+    Error = 4,
+}
+
+/// Metadata describing a metric event. This provides additional context to [`Recorder`](crate::Recorder), allowing for
+/// fine-grained filtering.
+/// 
+/// Contains the following:
+/// 
+/// - A [`target`](Metadata::target), specifying the part of the system where the metric event occurred. When
+/// initialized via the [metrics macro](metrics_macro), and left unspecified, this defaults to the module path the
```
tobz, line 33: Looks like this needs to be changed to something like `[metrics macros][metrics_macros]` and then add `[metrics_macros]: https://docs.rs/metrics/latest/metrics/#macros` at the bottom of the overall doc comment.

I don't think there's a way to do intradoc links with _just_ a crate name, but it's also cleaner/more intuitive, I think, to link the macros section of the `metrics` crate since that's where people will be consuming them from.

tobz (changes requested):
Left one little doc comment thing, and it looks like this needs to have `cargo fmt` run on it, but overall very close. 👍🏻

tobz:
I'm going to try and take a look, in a few days, at just finalizing the remaining requested changes myself since it's so close.

hlbarber:
"Rust / feature-check" is throwing a ` note: /usr/bin/ld: final link failed: No space left on device`, any suggestions?

tobz (approved):
LGTM.

The feature check CI step has been finicky lately with "out of disk space" errors cropping up a lot. I just check this stuff manually before cutting releases and it's usually fine, so just gonna ignore it for now.

Issue #369, opened 2023-05-11
Evaluate whether or not counter-related macros can be renamed/simplified.


tobz:
As mentioned in #338, it can be confusing when trying to figure out which counter macro to use for a given scenario, as we have `counter!`, `increment_counter!`, and `absolute_counter!`.

We should consider if there's any potential simplification we could apply. One form this proposed simplification could take is around exploring if one macro could serve both the "increment by N" and "increment by 1" use cases, rather than having them be distinct macros.

hlbarber:
I think it's important that others post APIs that they'd like here and we can compare the merits.

If you are unable to get a Rust playground proof-of-concept working I think it's still valuable to post and describe why you found it difficult. Maybe someone else will be able to make it work and then we can set up a boundary between the feasible and unfeasible APIs.

hlbarber:
> Here's a [working proposal](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6a64f0f94a76070722cee0a574a9162d) to get discussion started. Please "Tools -> Expand macros" and inspect.

One of the problems with this design, imo, is that `counter!("name", "labelA", "labelB")` is a bit confusing. Perhaps labels should be declared in the form `labels: $expr`, `labels: [$($keys),*]`, and `labels: [$(($key, $value)),*]`?

hlbarber:
Another route here is that we simply don't have any macro arguments for `+`, `-`, or `=` etc. Instead we just have
```rust
counter!("name").set(value);
counter!("name").add(value);
```

tobz:
Just talking out loud.... the pros/cons for the "use methods instead of individual operation macros":

### Pros

- Fewer macros overall!
- We can keep the metric name/target/level/labels handling fairly close to what it currently is: only the value for the operation would be moved as an argument to the operation method call.
- IDE support for documentation around actual methods/functions has always been vastly better IME than for macros, which would be good because I feel like documenting the nuances of the API, macros, etc, has always been a challenge.
- Gives us a way to easily support your idea above of the macro being able to serve both as a registration macro (`let foo = counter!(...)`) as well as an emission macro (`counter!(...).add(1)`) without having to thread the needle in terms of macro branch argument specificity.

### Cons

- Breaking change. We _are_ pre-1.0, but still a breaking change nonetheless.
- Somewhat unconventional, as the general pattern for macros is that they do everything within the call, or they're just used as a way to generate an expression, rather than ever used in a chained fashion. [1]

[...]

PR #386, opened 2023-08-25
migrate from procedural to declarative macros

hlbarber:
# Changes

- Remove `metrics-macros` crate
- Rewrite all existing macros using `macro_rules` within the `metrics` crate

The goal here is to not break users and to also maintain the "fast paths" taken by the procedural macro where literals cause `static` to be used.

~~This PR is intended as a place of discussion, once it is sufficiently mature I'll close it and reopen a polished PR.~~


hlbarber:
At random I've picked `register_counter` - from the initial experimentation it _feels_ possible. We're passing `cargo t` with a 2 exceptions.

metrics/src/macros.rs
```diff
@@ -0,0 +1,104 @@
...
+    };
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! key_var {
+    ($name: literal) => {
+        static METRIC_KEY: ::metrics::Key = ::metrics::Key::from_static_name($name);
+        &METRIC_KEY
+    };
```
tobz, line 50: Hmm, does this one need the double brackets to isolate the return value? 🤔

tobz (commented):
So far, so good... and in retrospect, much simpler than procedural macros. 😂 

Left some feedback.

metrics/src/macros.rs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            let metric_key = $crate::key_var! { $name, $($labels)? };
+            let metadata = $crate::metadata_var! {
+                $crate::default_target! { $($target)? },
+                $crate::default_level! { $($level)? }
+            };
+
+            ::metrics::recorder().register_counter(&metric_key, metadata)
+        }
+    };
```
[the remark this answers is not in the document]
hlbarber: My understanding (I'm also not well-versed) is that application will just iterate through the `macro_rules` branches, in order, and execute whatever matches first.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=86f118d95fa2f4c3071e698a8eda5b36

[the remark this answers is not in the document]
hlbarber: Here's an example of it a `expr` becoming a `literal` when passed through to an inner macro.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=80a7dfa8ec5e12c3e6e868c16b121a5c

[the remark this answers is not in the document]
tobz: Neat, another TIL.

tobz (commented):
Overall, this is seeming pretty reasonable and straightforward.

It does look like some tests are broken, though, specifically around how we're handling building the key, and the expectation that we can always borrow it at the callsite, even when it may already be borrowed.

hlbarber:
The following seem to fix the failing test cases while maintaining the fast paths where possible. Please tell me if this looks off in any way.

https://github.com/metrics-rs/metrics/pull/386/files/c356678f3bd0bab76ad6932a951374ca5a1e484d..5e82f3e41e5ad2034f03359286d2e1a01eb775cf

hlbarber:
That's the "counter" cross section of macros replaced. After `register_counter` it isn't really involved.

Given near perfect symmetry between the metric types I don't foresee any nasty obstructions when transitioning the other macros.

I'll check benchmarks shortly to check for any regressions.

hlbarber:
After restoring doc tests there's a bit more to do, will look into it tomorrow.

hlbarber:
https://github.com/metrics-rs/metrics/pull/386/commits/0a32d83394342889f0a5bdb8192b3d2c3216186d has made is considerably more branch-y, but still tractable imo. Rather than making use of `?` we must enumerate the combinations.

I think with some tinkering this can be compacted.

tobz:
> I _can_ push for compactness now, but given #369 is around the corner I can see an argument for postponing.

I'd say let's just get the logic correct now and deal with condensing later.

> I think with some tinkering this can be compacted. I think perhaps we can parse the arguments as one (or two?) `tt` and then pass them onto other branches - some prior art can be found in the `tracing` macros ([info](https://docs.rs/tracing/latest/tracing/macro.info.html) etc).

I'm not against the verbose-ness of having a branch for each possible permutation... so long as there's a viable path to, in the future, generalizing that macro logic across the different metric types. Basically, it'd be a small tragedy to just have to duplicate all of it over and over again when we do gauges and histograms... just ripe for introducing a subtle difference. 

Having one `register_metric` macro or something, for example, that we just call into and pass the metric type-appropriate idents, and so on... and it has all of the various branches? That'd be fine.

hlbarber:
> Having one `register_metric` macro or something, for example, that we just call into and pass the metric type-appropriate idents, and so on... and it has all of the various branches? That'd be fine.

A reason I haven't re-used `register_counter` in the current impl is that the branch, in `metric_macros::get_register_and_op_code` for `register_counter` ([found here](https://github.com/metrics-rs/metrics/blob/0193688dac4ca646dbe44620040c20b9abf9bf5e/metrics-macros/src/lib.rs#L289-L298)) uses `metrics::recorder` and the `counter`/`absolute_counter`/etc branch uses `metrics::try_recorder` ([found here](https://github.com/metrics-rs/metrics/blob/0193688dac4ca646dbe44620040c20b9abf9bf5e/metrics-macros/src/lib.rs#L281-L285)).

If I use `metrics::recorder` for the `counter` branch I get a regression on the "uninitialized" benchmarks - this is a little suprising to me, I'd expect the `NoopRecorder`, after optimizations, to be equal in performance to checking `Some(_) = try_recorder`.

tobz:
I'd say switch to `try_recorder` across the board.

The regression is surprising to me too, but also, thinking about it... I'm not sure I actually give a crap if it's like 5ns vs 20ns or whatever when a recorder isn't installed. If it ever matters enough to me, or someone else, well... we can figure something out then.

hlbarber:
Rest of the macros are migrated. I've DRY'd it up a little bit with the `describe` and `method` macros. I haven't taken the `try_recorder` change yet because it deduplicated less with `method` macro in place.

I played around a little bit with `$(args: tt),*`, trying something like

```rust
macro_rules counter {
    ($($args:tt),) => $crate::method(register_counter, increment, $($args),*)
}
```
similar with what we see in the `tracing` macros. This doesn't work because of the mix of delimiters we see in the macros (`,` and `=>`). You can see the exact problem here https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=def715a3d053a4b933ccb4a821260729. Perhaps this is something the next API will side step. I think there's also an argument to keeping the explicit branches in the "macro signatures" to serve as documentation.

`metrics-macro` is now empty. I'm guessing I should delete the crate from the workspace entirely so that we don't release an empty crate?

tobz:
> Perhaps this is something the next API will side step. I think there's also an argument to keeping the explicit branches in the "macro signatures" to serve as documentation.

I think things are fine as is. Good to see your example as it makes it clear, but in seeing it and seeing what you chose to do instead, it makes me think that making it perfectly DRY isn't as big a readability win as I thought it might be anyways.

> `metrics-macros` is now empty. I'm guessing I should delete the crate from the workspace entirely so that we don't release an empty crate?

Yessss, nuke it from orbit. 👨🏻‍🚀

metrics/src/macros.rs
```diff
@@ -0,0 +1,809 @@
...
+#[doc(hidden)]
+#[macro_export]
+macro_rules! describe {
+    ($method:ident, $name:expr, $unit:expr, $description:expr) => {{
+        if let ::core::option::Option::Some(recorder) = $crate::try_recorder() {
+            recorder.$method($name.into(), ::core::option::Option::Some($unit), $description.into())
+        }
+    }};
+    ($method:ident, $name:expr, $description:expr) => {{
+        if let Some(recorder) = $crate::try_recorder() {
```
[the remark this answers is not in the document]
hlbarber: It does, whoops. I'll try write a test involving `#[no_implicit_prelude]` so that me and others wont make this mistake again.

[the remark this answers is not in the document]
hlbarber: https://github.com/metrics-rs/metrics/pull/386/commits/a901339b4c122d0c2486ba7e44031b80e097c786

I've slipped `#![no_implicit_prelude]` into the doc tests.

I think it might be better to write a dedicate test (or even just an example) where we invoke all combinations of all macros with `#![no_implicit_prelude]` and we don't import things like `use std::convert::From` which weakens the test. I can do this in a follow up PR unless you think its important now.

tobz:
I plan on taking a final look at this over the next few days.

hlbarber:
I can transplant this into a formal PR with a proper description now if you'd like?

tobz:
Nah, no worries. At most, feel free to update the PR title and description and mark it as ready to review and all of that... but no further cleanup/tidying required.

metrics/src/macros.rs
```diff
@@ -0,0 +1,870 @@
...
+    ($name: literal) => {{
+        static METRIC_KEY: $crate::Key = $crate::Key::from_static_name($name);
+        &METRIC_KEY
+    }};
+    ($name:expr) => {
+        $crate::Key::from_name($name)
+    };
+    ($name:literal, $($label_key:literal => $label_value:literal),*) => {{
+        static LABELS: [$crate::Label; $crate::count!($($label_key)*)] = [
+            $($crate::Label::from_static_parts(&$label_key, &$label_value)),*
```
tobz, line 34: I don't think the label name/value have to be referenced here since they're already literals?

metrics/src/macros.rs
```diff
@@ -0,0 +1,870 @@
...
+        let metric_key = $crate::key_var!($name $(, $label_key $(=> $label_value)?)*);
+        let metadata = $crate::metadata_var!($target, $level);
+
+        $crate::recorder().register_counter(&metric_key, metadata)
+    }};
+    (target: $target:expr, $name:expr $(, $label_key:expr $(=> $label_value:expr)?)* $(,)?) => {
+        $crate::register_counter!(target: $target, level: $crate::Level::INFO, $name $(, $label_key $(=> $label_value)?)*)
+    };
+    (level: $level:expr, $name:expr $(, $label_key:expr $(=> $label_value:expr)?)* $(,)?) => {
+        $crate::register_counter!(target: module_path!(), level: $level, $name $(, $label_key$(=> $label_value)?)*)
```
tobz, line 115: Should references to `module_path!()` also be qualified paths like you did above with `::std::vec!`?

metrics/src/macros.rs
```diff
@@ -0,0 +1,870 @@
...
+        let metric_key = $crate::key_var!($name $(, $label_key $(=> $label_value)?)*);
+        let metadata = $crate::metadata_var!($target, $level);
+
+        $crate::recorder().register_gauge(&metric_key, metadata)
+    }};
+    (target: $target:expr, $name:expr $(, $label_key:expr $(=> $label_value:expr)?)* $(,)?) => {
+        $crate::register_gauge!(target: $target, level: $crate::Level::INFO, $name $(, $label_key $(=> $label_value)?)*)
+    };
+    (level: $level:expr, $name:expr $(, $label_key:expr $(=> $label_value:expr)?)* $(,)?) => {
+        $crate::register_gauge!(target: module_path!(), level: $level, $name $(, $label_key$(=> $label_value)?)*)
```
tobz, line 182: Looks like at least for clarity's sake, this should be `$(, $label_key $(=> $label_value)?)*` instead of `$(, $label_key$(=> $label_value)?)*`. Same with the other similar callsites.

metrics/src/macros.rs
```diff
@@ -0,0 +1,870 @@
...
+    }
+}
+
+/// Registers a counter.
+///
+/// Counters represent a single monotonic value, which means the value can only be incremented, not
+/// decremented, and always starts out with an initial value of zero.
+///
+/// Metrics can be registered, which provides a handle to directly update that metric.  For
+/// counters, [`Counter`](crate:Counter) is provided which can be incremented or set to an absolute value.
```
tobz, line 62: This should be `crate::Counter`.

tobz (changes requested):
Overall, this looks really good and most of the comments are nits. There's one comment about a broken rustdoc link that needs to be fixed, however.

metrics/src/macros.rs
hlbarber: Oddly this `module_path` seems to still be imported when `no_implicit_prelude` is on, despite other macros like `println` not working.

As `module_path` can be shadowed also definitely should yeah.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=612b6e150612da97bd5d4dbc221ed40b

hlbarber: 11b36bc8c00c91d8b15db9a4bcda67ebe74d6549

metrics/src/macros.rs
hlbarber: 02f230f25d4b6b3ee0074fd8ee0fbd60a075dd26

metrics/src/macros.rs
hlbarber: d70bb0e6e58f01f8c5b224a98b511000db182261

metrics/src/macros.rs
hlbarber: 28194901f7a6ed6e995248f87e8577536adc504a

metrics/src/macros.rs
hlbarber: ah `no_implicit_prelude` does not disable https://doc.rust-lang.org/reference/names/preludes.html#language-prelude

hlbarber:
On my machine `cargo bench`, on `main` vs this branch, shows no obvious regressions - worth double checking on your machine though.

metrics/src/macros.rs
tobz: Makes sense. 👍🏻

tobz:
Released as `metrics@v0.22.0`.

PR #390, opened 2023-09-14
Expand lifetime of KeyName as_str

miguelff:
A `SharedString` is valid during the duration of the program, once built. This PR extends the lifetime of the string slice returned by `as_str` to avoid needing to clone the key name.


miguelff:
Ohhhh! right, thanks for explaining. This is a part of the ownership model that is still not fully clear in my head. Please excuse the noise.

tobz:
No worries at all. 👍🏻

Issue #391, opened 2023-09-15
:bug: [prometheus-exporter] Push-gateway does not support https or auth


qingwen220:
:bug:  Push-gateway in metric-exporter-prometheus does not support HTTPS or auth even though user & password are required to be provided. 

Because the hyper client only supports HTTP protocol in builder.rs.

PR #392, opened 2023-09-15
add https support in Prometheus gateway

qingwen220:
Fixes: #391 

Reproduce: any URL with https

### DIFF
--- a/metrics-exporter-prometheus/src/builder.rs
+++ b/metrics-exporter-prometheus/src/builder.rs
@@ -29,6 +29,9 @@ use hyper::{
     http::HeaderValue,
     Method, Request, Uri,
 };
+use hyper::client::{HttpConnector};
+use hyper::http::uri::Scheme;
+use hyper_tls::HttpsConnector;
 
 use indexmap::IndexMap;
 #[cfg(feature = "http-listener")]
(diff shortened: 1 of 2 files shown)


metrics-exporter-prometheus/src/builder.rs
```diff
@@ -566,6 +574,20 @@ fn basic_auth(username: &str, password: Option<&str>) -> HeaderValue {
...
+
+fn create_gateway_client(endpoint: &Uri) -> HttpCli {
+    if endpoint.scheme() == Some(&Scheme::HTTPS) {
+        let https = HttpsConnector::new();
+        HttpCli::HttpsClient(Client::builder().build::<_, hyper::Body>(https))
+    } else {
+        HttpCli::HttpClient(Client::new())
+    }
+}
+
```
tobz, line 590: We should always use the `HttpsConnector`-based version, regardless of the scheme. [`HttpsConnector::new`](https://docs.rs/hyper-tls/latest/hyper_tls/struct.HttpsConnector.html#method.new) mentions that it will handle HTTP vs HTTPS properly, so we can avoid all this extra code.

tobz (changes requested):

metrics-exporter-prometheus/src/builder.rs
qingwen220: @tobz  You're right. I've updated the code after double-checked the implementation in hyper-tls.

tobz:
Released as `metrics-exporter-prometheus@v0.13.0`.

PR #393, opened 2023-09-21
Make `Unit` methods return `&'static str` where possible

david-perez:


tobz:
Released as `metrics@v0.22.0`.

PR #394, opened 2023-10-01
remove macros, retaining only `register_*` and `describe_*` macros

hlbarber:
## Motivation

Simplification of the existing macro API

https://github.com/metrics-rs/metrics/issues/369

## Changes

- Remove `increment_counter`, `absolute_counter`, `increment_gauge`, `decrement_gauge`, `gauge`, `histogram`.
- Rename `register_counter`, `register_gauge`, and `record_histogram` to `counter`, `gauge`, and `histogram` respectively.
- Update documentation.

### DIFF
--- a/metrics/examples/basic.rs
+++ b/metrics/examples/basic.rs
@@ -8,9 +8,8 @@
 use std::sync::Arc;
 
 use metrics::{
-    absolute_counter, counter, decrement_gauge, describe_counter, describe_gauge,
-    describe_histogram, gauge, histogram, increment_counter, increment_gauge, register_counter,
-    register_gauge, register_histogram, KeyName, Metadata, SharedString,
+    counter, describe_counter, describe_gauge, describe_histogram, gauge, histogram, KeyName,
+    Metadata, SharedString,
 };
 use metrics::{Counter, CounterFn, Gauge, GaugeFn, Histogram, HistogramFn, Key, Recorder, Unit};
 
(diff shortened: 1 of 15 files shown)


LukeMathWalker:
What prevents the `describe_*` macro family from being converted into a set of methods on the respective metric types?
Having a single macro per metric type would be quite interesting—it would entirely remove ambiguity when it comes to "what macro do I need".

hlbarber:
> What prevents the `describe_*` macro family from being converted into a set of methods on the respective metric types?

Yeah removing the family of `describe_*` macros would be super nice and adding them as just an extra methods onto the handles would be super ergonomic.

The obstruction I see here is that `Recorder::describe_*` takes a `KeyName` ([example](https://docs.rs/metrics/latest/metrics/trait.Recorder.html#tymethod.describe_counter)) whereas `Recorder::register_*`takes `Key` ([example](https://docs.rs/metrics/latest/metrics/trait.Recorder.html#tymethod.register_counter)), meaning that `register_*` requires labels to be specified whereas `describe_*` doesn't. Given that the current `counter!` API accepts label arguments it might be confusing that `counter!("name", a => b).describe(...)` disregards `a => b`.

What you could imagine is moving the `label` arguments from `counter!` and putting them into the `increment` method - `counter!("name").increment(3, labels)`. Or maybe `counter!("name").labels(labels).increment(3)`. This would mean that we forgo some of the performance conscious "specialization" taken by the macro - would have to benchmark to check for regressions.

tobz:
I have Thoughts (tm) about all of this, and it's probably worth a new issue to avoid polluting this PR.

metrics/examples/basic.rs
```diff
@@ -119,54 +118,59 @@ fn main() {
...
-    increment_counter!("requests_processed", "request_type" => "admin");
-    increment_counter!("requests_processed", "request_type" => "admin", "server" => server_name.clone());
-    increment_counter!("requests_processed", common_labels);
-
-    absolute_counter!("bytes_sent", 64);
-    absolute_counter!("bytes_sent", 64, "listener" => "frontend");
-    absolute_counter!("bytes_sent", 64, "listener" => "frontend", "server" => server_name.clone());
-    absolute_counter!("bytes_sent", 64, common_labels);
+    counter!("bytes_sent").increment(64);
+    counter!("bytes_sent",  "listener" => "frontend").increment(64);
```
tobz, line 138:
```suggestion
    counter!("bytes_sent", "listener" => "frontend").increment(64);
```

tobz (approved):
Overall all, this looks good to me. 👍🏻 

Left one _super_ small nit.

tobz:
Just gonna merge since the feature checks are known to go boom on the free GHA runners... something I need to eventually take care of.

kornelski:
This change wasn't mentioned in the changelog.

tobz:
Ah yeah, it does look like I forgot to add a link to RELEASES.md in the change log for this specific change.

Easy enough to backport. Thanks for reporting. 👍

PR #408, opened 2023-11-06
Add support for `tracing::Span::record`ed fields in `metrics-tracing-context`

zohnannor:
I wondered for the reason of why [`record`ed fields are not added to the `Labels`](https://docs.rs/metrics-tracing-context/0.14.0/metrics_tracing_context/index.html#lack-of-dynamism) and found none. Implemented it as just `Label::extend_from_labels` call in a `Layer::on_record` method. For the tests I am not sure, maybe I missed some corner case, but I can't see such a case.

P.S. Should I ignore falling CI checks?

### DIFF
--- a/metrics-tracing-context/Cargo.toml
+++ b/metrics-tracing-context/Cargo.toml
@@ -32,6 +32,7 @@ itoa = { version = "1", default-features = false }
 metrics = { version = "^0.21", path = "../metrics" }
 metrics-util = { version = "^0.15", path = "../metrics-util" }
 lockfree-object-pool = { version = "0.1.3", default-features = false }
+indexmap = "2.1"
 once_cell = { version = "1", default-features = false, features = ["std"] }
 tracing = { version = "0.1.29", default-features = false }
 tracing-core = { version = "0.1.21", default-features = false }
--- a/metrics-tracing-context/Cargo.toml
+++ b/metrics-tracing-context/Cargo.toml
@@ -37,10 +37,12 @@ once_cell = { version = "1", default-features = false, features = ["std"] }
 tracing = { version = "0.1.29", default-features = false }
 tracing-core = { version = "0.1.21", default-features = false }
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["std"] }
+tracing-test = "0.2.4"
 
 [dev-dependencies]
 criterion = { version = "=0.3.3", default-features = false }
 parking_lot = { version = "0.12.1", default-features = false }
 tracing = { version = "0.1.29", default-features = false, features = ["std"] }
-tracing-subscriber = { version = "0.3.1", default-features = false, features = ["registry"] }
+tracing-subscriber = { version = "0.3.1", default-features = false, features = ["registry", "fmt"] }
 pretty_assertions = "*"
+itertools = "*"
--- a/metrics-tracing-context/Cargo.toml
+++ b/metrics-tracing-context/Cargo.toml
@@ -43,3 +43,4 @@ criterion = { version = "=0.3.3", default-features = false }
 parking_lot = { version = "0.12.1", default-features = false }
 tracing = { version = "0.1.29", default-features = false, features = ["std"] }
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["registry"] }
+pretty_assertions = "*"
--- a/metrics-tracing-context/Cargo.toml
+++ b/metrics-tracing-context/Cargo.toml
@@ -37,7 +37,6 @@ once_cell = { version = "1", default-features = false, features = ["std"] }
 tracing = { version = "0.1.29", default-features = false }
 tracing-core = { version = "0.1.21", default-features = false }
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["std"] }
-tracing-test = "0.2.4"
 
 [dev-dependencies]
 criterion = { version = "=0.3.3", default-features = false }
@@ -45,4 +44,4 @@ parking_lot = { version = "0.12.1", default-features = false }
 tracing = { version = "0.1.29", default-features = false, features = ["std"] }
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["registry", "fmt"] }
 pretty_assertions = "*"
-itertools = "*"
+itertools = { version = "0.12.0", default-features = false, features = ["use_std"] }
--- a/metrics-tracing-context/Cargo.toml
+++ b/metrics-tracing-context/Cargo.toml
@@ -32,7 +32,7 @@ itoa = { version = "1", default-features = false }
 metrics = { version = "^0.21", path = "../metrics" }
 metrics-util = { version = "^0.15", path = "../metrics-util" }
 lockfree-object-pool = { version = "0.1.3", default-features = false }
-indexmap = "2.1"
+indexmap = { version = "2.1", default-features = false, features = ["std"] }
 once_cell = { version = "1", default-features = false, features = ["std"] }
 tracing = { version = "0.1.29", default-features = false }
 tracing-core = { version = "0.1.21", default-features = false }
--- a/metrics-tracing-context/Cargo.toml
+++ b/metrics-tracing-context/Cargo.toml
@@ -42,6 +42,5 @@ tracing-subscriber = { version = "0.3.1", default-features = false, features = [
 criterion = { version = "=0.3.3", default-features = false }
 parking_lot = { version = "0.12.1", default-features = false }
 tracing = { version = "0.1.29", default-features = false, features = ["std"] }
-tracing-subscriber = { version = "0.3.1", default-features = false, features = ["registry", "fmt"] }
-pretty_assertions = "*"
+tracing-subscriber = { version = "0.3.1", default-features = false, features = ["registry"] }
 itertools = { version = "0.12.0", default-features = false, features = ["use_std"] }
--- a/metrics-tracing-context/src/lib.rs
+++ b/metrics-tracing-context/src/lib.rs
@@ -110,8 +110,8 @@ pub mod label_filter;
 mod tracing_integration;
 
 pub use label_filter::LabelFilter;
+use tracing_integration::Map;
 pub use tracing_integration::{Labels, MetricsLayer};
-use tracing_integration::{Map, WithContext};
 
 /// [`TracingContextLayer`] provides an implementation of a [`Layer`] for [`TracingContext`].
 pub struct TracingContextLayer<F> {
@@ -176,7 +176,7 @@ where
         tracing::dispatcher::get_default(|dispatch| {
             let current = dispatch.current_span();
             let id = current.id()?;
-            let ctx = dispatch.downcast_ref::<WithContext>()?;
+            let ctx = dispatch.downcast_ref::<MetricsLayer>()?;
 
             let mut f = |mut span_labels: Map| {
                 (!span_labels.is_empty()).then(|| {
--- a/metrics-tracing-context/src/lib.rs
+++ b/metrics-tracing-context/src/lib.rs
@@ -186,7 +186,7 @@ where
 
                     let labels = span_labels
                         .into_iter()
-                        .map(|(key, value)| Label::new(key.clone(), value.clone()))
+                        .map(|(key, value)| Label::new(key, value))
                         .filter(|label| self.label_filter.should_include_label(&name, label))
                         .collect::<Vec<_>>();
 
--- a/metrics-tracing-context/src/lib.rs
+++ b/metrics-tracing-context/src/lib.rs
@@ -55,19 +55,18 @@
 //!
 //! # Implementation
 //!
-//! The integration layer works by capturing all fields present when a span is created and storing
-//! them as an extension to the span.  If a metric is emitted while a span is entered, we check that
-//! span to see if it has any fields in the extension data, and if it does, we add those fields as
-//! labels to the metric key.
+//! The integration layer works by capturing all fields present when a span is
+//! created or then new fields are recorded afterward, and storing them as an
+//! extension to the span.  If a metric is emitted while a span is entered, we
+//! check that span to see if it has any fields in the extension data, and if
+//! it does, we add those fields as labels to the metric key.
 //!
-//! There are two important behaviors to be aware of:
-//! - we only capture the fields present when the span is created
-//! - we store all fields that a span has, including the fields of its parent span(s)
+//! Be aware that we store all fields that a span has, including the fields of its parent span(s).
 //!
-//! ## Lack of dynamism
+//! ## Support for dynamism
 //!
 //! This means that if you use [`Span::record`][tracing::Span::record] to add fields to a span after
-//! it has been created, those fields will not be captured and added to your metric key.
+//! it has been created, those fields will be captured and added to your metric key.
 //!
 //! ## Span fields and ancestry
 //!
--- a/metrics-tracing-context/src/lib.rs
+++ b/metrics-tracing-context/src/lib.rs
@@ -101,7 +101,9 @@
 #![deny(missing_docs)]
 #![cfg_attr(docsrs, feature(doc_cfg), deny(rustdoc::broken_intra_doc_links))]
 
-use metrics::{Counter, Gauge, Histogram, Key, KeyName, Metadata, Recorder, SharedString, Unit};
+use metrics::{
+    Counter, Gauge, Histogram, Key, KeyName, Label, Metadata, Recorder, SharedString, Unit,
+};
 use metrics_util::layers::Layer;
 
 pub mod label_filter;
@@ -173,32 +175,27 @@ where
         // doing the iteration would likely exceed the cost of simply constructing the new key.
         tracing::dispatcher::get_default(|dispatch| {
             let current = dispatch.current_span();
-            if let Some(id) = current.id() {
-                // We're currently within a live tracing span, so see if we have an available
-                // metrics context to grab any fields/labels out of.
-                if let Some(ctx) = dispatch.downcast_ref::<WithContext>() {
-                    let mut f = |new_labels: &Map| {
-                        (!new_labels.is_empty()).then(|| {
-                            let (name, mut labels) = key.clone().into_parts();
+            let id = current.id()?;
+            let ctx = dispatch.downcast_ref::<WithContext>()?;
 
-                            let filtered_labels = new_labels
-                                .values()
-                                .filter(|label| {
-                                    self.label_filter.should_include_label(&name, label)
-                                })
-                                .cloned();
-                            labels.extend(filtered_labels);
+            let mut f = |mut span_labels: Map| {
+                (!span_labels.is_empty()).then(|| {
+                    let (name, labels) = key.clone().into_parts();
 
-                            Key::from_parts(name, labels)
-                        })
-                    };
+                    span_labels.extend(labels.into_iter().map(Label::into_parts));
 
-                    // Pull in the span's fields/labels if they exist.
-                    return ctx.with_labels(dispatch, id, &mut f);
-                }
-            }
+                    let labels = span_labels
+                        .into_iter()
+                        .map(|(key, value)| Label::new(key.clone(), value.clone()))
+                        .filter(|label| self.label_filter.should_include_label(&name, label))
+                        .collect::<Vec<_>>();
 
-            None
+                    Key::from_parts(name, labels)
+                })
+            };
+
+            // Pull in the span's fields/labels if they exist.
+            ctx.with_labels(dispatch, id, &mut f)
         })
     }
 }
--- a/metrics-tracing-context/src/lib.rs
+++ b/metrics-tracing-context/src/lib.rs
@@ -55,18 +55,17 @@
 //!
 //! # Implementation
 //!
-//! The integration layer works by capturing all fields present when a span is
-//! created or then new fields are recorded afterward, and storing them as an
-//! extension to the span.  If a metric is emitted while a span is entered, we
-//! check that span to see if it has any fields in the extension data, and if
-//! it does, we add those fields as labels to the metric key.
+//! The integration layer works by capturing all fields present when a span is created or then
+//! new fields are recorded afterward, and storing them as an extension to the span.  If a metric
+//! is emitted while a span is entered, we check that span to see if it has any fields in the
+//! extension data, and if it does, we add those fields as labels to the metric key.
 //!
 //! Be aware that we store all fields that a span has, including the fields of its parent span(s).
 //!
 //! ## Support for dynamism
 //!
-//! This means that if you use [`Span::record`][tracing::Span::record] to add fields to a span after
-//! it has been created, those fields will be captured and added to your metric key.
+//! If you use [`Span::record`][tracing::Span::record] to add fields to a span after it has been
+//! created, those fields will be captured and added to your metric key.
 //!
 //! ## Span fields and ancestry
 //!
--- a/metrics-tracing-context/src/lib.rs
+++ b/metrics-tracing-context/src/lib.rs
@@ -55,19 +55,25 @@
 //!
 //! # Implementation
 //!
-//! The integration layer works by capturing all fields present when a span is created or then
-//! new fields are recorded afterward, and storing them as an extension to the span.  If a metric
-//! is emitted while a span is entered, we check that span to see if it has any fields in the
-//! extension data, and if it does, we add those fields as labels to the metric key.
+//! The integration layer works by capturing all fields that are present when a span is created,
+//! as well as fields recorded after the fact, and storing them as an extension to the span. If
+//! a metric is emitted while a span is entered, any fields captured for that span will be added
+//! to the metric as additional labels.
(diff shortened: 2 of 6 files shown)


tobz (changes requested):
I still need to read over the documentation changes, but left at least one bit of feedback on a required change.

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -146,6 +145,24 @@ where
...
 
+    fn on_record(&self, id: &Id, values: &Record<'_>, cx: Context<'_, S>) {
+        let span = cx.span(id).expect("span must already exist!");
+        let mut labels = Labels::from_record(values);
+
+        if let Some(parent) = span.parent() {
+            if let Some(parent_labels) = parent.extensions().get::<Labels>() {
+                labels.extend_from_labels(parent_labels);
+            }
+        }
```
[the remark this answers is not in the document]
zohnannor: Right! Overlooked this bit, thank you. Will fix it when I can.

metrics-tracing-context/tests/integration.rs
```diff
@@ -110,7 +117,137 @@ fn test_basic_functionality() {
...
+
+        assert_eq!(snapshot, snapshots);
+    }
+}
+
+#[test]
+fn test_rerecord() {
+    static USER_ID_42: &[Label] = &[Label::from_static_parts("user.id", "42")];
+    static USER_ID_123: &[Label] =
+        &[Label::from_static_parts("user.id", "42"), Label::from_static_parts("user.id", "123")];
```
zohnannor, line 213: If the field is `record`ed more than one time, each time emitting a metric, we would record every seen value. Is that a concern? Probably should be documented somewhere?

zohnannor: `tracing`, in one (or many) of their `Subscriber` `Layer`s, currently has the same problem, here's the open issue: https://github.com/tokio-rs/tracing/issues/1372#issuecomment-1709274609.

zohnannor: One obvious downside is something like this:
```rust
let span = tracing::info_span!("loop", item = field::Empty);
let _g = span.enter();
for i in foo() {
	span.record("item", i);
	metrics::counter!("counter").increment(1);
}
```
Would cause us to extend `Labels` with each iteration.

zohnannor: We could store `Labels` as `pub struct Labels(pub LinearOwnedReusable<IndexMap<SharedString, Label>>);`. That would be a breaking change for two reasons: 1) `Labels` are `pub` (despite the `doc(hidden)`); 2) Nested spans  with same field name would overwrite each-other, as opposed to current behavior (`test_nested_spans` test).

tobz: Yeah, this is all a little bit tricky.

I don't love the idea of having to also store a map for every span, but there's obviously got to be some deduplication logic _somewhere_. I'm just trying to think through if it should live in the layer logic (so that we amortize it across all `record` calls, etc) or if it should happen when we want to collect/build the final set of labels before creating the new metric key. 🤔

zohnannor: Let's evaluate the trade-off if storing the map is undesirable:

1. Deduplicate in `on_record`:
    `+`: we store only required labels, saving on allocations.
	`-`: possibly a lot of extra operations on a "happy path" when user does not re-record the fields.
	`+`: key creation becomes free
2. Deduplicate on key creation:
	`+`: it would be the only time we need to deduplicate labels.
	`-`: we store all previous `record`ed values for a label, extending the underlying `Vec` each time.

tobz: Thinking more about it, your point about extending the labels on each iteration of the loop in your example got me to thinking: even if we were willing to eat the cost of duplicate entries... the more risky part is that we could end up extending that array of labels in an infinite/unbounded fashion. For example, if your loop was an asynchronous iterator, and recorded a field every single time, that stream may never end, etc.

I'd say let's go with the `IndexMap<K, V>`-based approach.

metrics-tracing-context/Cargo.toml
```diff
@@ -32,6 +32,7 @@ itoa = { version = "1", default-features = false }
 metrics = { version = "^0.21", path = "../metrics" }
 metrics-util = { version = "^0.15", path = "../metrics-util" }
 lockfree-object-pool = { version = "0.1.3", default-features = false }
+indexmap = "2.1"
```
tobz, line 35: Can you add `default-features = false` here? It's fine if we need to re-enable a specific feature, but we don't want to include default features automatically.

metrics-tracing-context/src/lib.rs
```diff
@@ -55,19 +55,19 @@
...
 //! # Implementation
 //!
-//! The integration layer works by capturing all fields present when a span is created and storing
-//! them as an extension to the span.  If a metric is emitted while a span is entered, we check that
-//! span to see if it has any fields in the extension data, and if it does, we add those fields as
-//! labels to the metric key.
+//! The integration layer works by capturing all fields present when a span is created or then
+//! new fields are recorded afterward, and storing them as an extension to the span.  If a metric
+//! is emitted while a span is entered, we check that span to see if it has any fields in the
+//! extension data, and if it does, we add those fields as labels to the metric key.
```
tobz, line 61:
```suggestion
//! The integration layer works by capturing all fields that are present when a span is created,
//! as well as fields recorded after the fact, and storing them as an extension to the span. If
//! a metric is emitted while a span is entered, any fields captured for that span will be added
//! to the metric as additional labels.
```

metrics-tracing-context/src/lib.rs
```diff
@@ -55,19 +55,19 @@
...
 //!
-//! There are two important behaviors to be aware of:
-//! - we only capture the fields present when the span is created
-//! - we store all fields that a span has, including the fields of its parent span(s)
+//! Be aware that we store all fields that a span has, including the fields of its parent span(s).
```
tobz, line 63:
```suggestion
//! Be aware that we capture all recursively capture the fields of a span, including fields from
//! parent spans, and use them when generating metric labels. This means that if a metric is being
//! emitted in span B, which is a child of span A, and span A has field X, and span B has field Y,
//! then the metric labels will include both field X and Y. This applies regardless of how many
//! nested spans are currently entered.
```

metrics-tracing-context/src/lib.rs
```diff
@@ -55,19 +55,19 @@
...
 //!
-//! ## Lack of dynamism
+//! ## Support for dynamism
```
tobz, line 65:
```suggestion
//! ## Duplicate span fields
```

metrics-tracing-context/src/lib.rs
```diff
@@ -55,19 +55,19 @@
...
 //!
-//! This means that if you use [`Span::record`][tracing::Span::record] to add fields to a span after
-//! it has been created, those fields will not be captured and added to your metric key.
+//! If you use [`Span::record`][tracing::Span::record] to add fields to a span after it has been
+//! created, those fields will be captured and added to your metric key.  Multiple records of the
+//! same field would overwrite it, leaving the most recent value.  This way you can change any of
+//! the metrics keys coming from span fields.
```
tobz, line 70:
```suggestion
//! When span fields are captured, they are deduplicated such that only the most recent value is kept.
//! For merging parent span fields into the current span fields, the fields from the current span have
//! the highest priority. Additionally, when using [`Span::record`][tracing::Span::record] to add fields
//! to a span after it has been created, the same behavior applies. This means that recording a field
//! multiple times only keeps the most recently recorded value, including if a field was already present
//! from a parent span and is then recorded dynamically in the current span.
```

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -35,46 +44,45 @@ impl Default for Labels {
 impl Visit for Labels {
     fn record_str(&mut self, field: &Field, value: &str) {
         let label = Label::new(field.name(), value.to_string());
-        self.0.push(label);
+        self.0.insert(field.name().into(), label);
```
tobz, line 47: We should either:

- use `&'static str' as the map key type since we don't use the key for anything other than deduplicating
- keep the key as `SharedString` but then change the map value type to `SharedString` and construct the `Label` from the key/value in `enhance_key`

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -123,8 +131,7 @@ where
             .expect("subscriber should downcast to expected type; this is a bug!");
         let span = subscriber.span(id).expect("registry should have a span for the current ID");
 
-        let result =
-            if let Some(labels) = span.extensions().get::<Labels>() { f(labels) } else { None };
+        let result = span.extensions().get::<Labels>().and_then(f);
         result
```
tobz, line 135: We should just do:

```suggestion
span.extensions().get::<Labels>().and_then(f)
```

I'm not sure why I/whoever originally split it across two lines. 🤔

tobz (changes requested):
Overall, this is looking good and just needs a little more polish. I still haven't looked fully at the unit tests or benchmarks, since I just wanted to make sure any relevant code changes are in place before thinking about that stuff.

metrics-tracing-context/src/tracing_integration.rs
zohnannor: Went with the first option

metrics-tracing-context/src/tracing_integration.rs
zohnannor: How did I miss it :smile:

zohnannor: Oh, I know why:

```rust
error[E0597]: `span` does not live long enough
   --> metrics-tracing-context/src/tracing_integration.rs:134:9
    |
132 |         let span = subscriber.span(id).expect("registry should have a span for the current ID");
    |             ---- binding `span` declared here
133 |
134 |         span.extensions().get::<Labels>().and_then(f)
    |         ^^^^-------------
    |         |
    |         borrowed value does not live long enough
    |         a temporary with access to the borrow is created here ...
135 |     }
    |     -
    |     |
    |     `span` dropped here while still borrowed
    |     ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `Extensions<'_>`
    |
    = note: the temporary is part of an expression at the end of a block;
            consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
    |
134 |         let x = span.extensions().get::<Labels>().and_then(f); x
    |         +++++++                                              +++

```

That's weird that by just assigning it to a binding the error is resolved.

So either leave as is, or:

```rust
        let ext = span.extensions();
        ext.get::<Labels>().and_then(f)
```

I would go with the latter as it is more ... easier to get? And is a common pattern, because `.extensions()` is the source of the temporary in the chained methods expression.

tobz: Oh, well that's annoying. Yeah, I agree with your idea; let's go with that approach.

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -146,6 +153,18 @@ where
...
+
+        let ext = &mut span.extensions_mut();
+        if let Some(existing) = ext.get_mut::<Labels>() {
+            existing.extend_from_labels(&labels);
+        } else {
+            ext.insert(labels);
+        }
+    }
+
     unsafe fn downcast_raw(&self, id: TypeId) -> Option<*const ()> {
```
zohnannor, line 168: By the way, there is a way to get rid of this `unsafe` implementation. If you're interested, I can provide more info, or even make a PR.

tobz: There's not, as it's part of the [`Subscriber`](https://docs.rs/tracing-core/latest/tracing_core/trait.Subscriber.html#method.downcast_raw) trait itself.

tobz (changes requested):
One more small functional change we need to make, then I'll do the final review over the tests.

Appreciate your patience so far on this as my time has been a bit limited lately. 😓

metrics-tracing-context/src/tracing_integration.rs
zohnannor: What I mean is it's not needed for functionality provided by this crate. There is, and I already used this.

Here's the general idea:

```rust
use tracing::{span, Dispatch, Subscriber};
use tracing_subscriber::{
    layer::SubscriberExt as _, registry::LookupSpan, util::SubscriberInitExt as _, Layer,
};

fn main() {
    tracing_subscriber::registry()
        .with(MySubscriber::new())
        .init();

    tracing::info_span!("parent").in_scope(|| {
        tracing::info_span!("child").in_scope(|| {
            assert_eq!(get_parent_name().as_deref(), Some("parent"));
        });
    });
}

fn get_parent_name() -> Option<String> {
    tracing::Span::current()
        .with_subscriber(|(id, dispatch)| {
            dispatch
                .downcast_ref::<MySubscriber>()?
                .get_parent_name(dispatch, id)
        })
        .flatten()
}

struct MySubscriber {
    get_parent_name: Option<fn(dispatch: &Dispatch, id: &span::Id) -> Option<String>>,
}

impl MySubscriber {
    fn new() -> Self {
        Self {
            get_parent_name: None,
        }
    }

    fn get_parent_name(&self, dispatch: &Dispatch, id: &span::Id) -> Option<String> {
        (self.get_parent_name?)(dispatch, id)
    }
}

impl<S> Layer<S> for MySubscriber
where
    S: Subscriber + for<'span> LookupSpan<'span>,
{
    fn on_layer(&mut self, _: &mut S) {
        self.get_parent_name = Some(|dispatch: &Dispatch, id: &span::Id| {
            let sub = dispatch.downcast_ref::<S>().unwrap();
            let span = sub.span(id).unwrap();

            span.parent().map(|p| p.name().to_owned())

            // Instead of this 👆, do some logic with callsite-provided data and spanref
        });
    }
}
```

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -1,28 +1,37 @@
...
 
 impl Labels {
     pub(crate) fn extend_from_labels(&mut self, other: &Labels) {
-        self.0.extend_from_slice(other.as_ref());
+        let new_len = cmp::max(self.as_ref().len(), other.as_ref().len());
+        let additional = new_len - self.as_ref().len();
+        self.0.reserve(additional);
+        for (k, v) in other.as_ref() {
+            self.0.insert(k, v.clone());
+        }
```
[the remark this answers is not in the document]
zohnannor: So, the opposite of what's described here? https://github.com/metrics-rs/metrics/pull/408/files#diff-a16803449fe3229b496af705d5d434d1dee993cbd2606a6fb04f3a6a645a9955R71-R76 I thought it would be correct to overwrite them. Or do you mean this:

```rust
tracing::info_span!("span", field = 1).in_scope(|| {
	metrics::counter(
		"counter",
		"field" => 2,
	).increment(1); // would increment `counter{field=1}`, because value `2` is overwritten by the span
});
```

In that case, yeah, this should be done differently then.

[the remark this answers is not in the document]
zohnannor: This is tricky, as we want to keep only the "most recent" value from spans, but discard it entirely if it clashes with one of the labels originally used...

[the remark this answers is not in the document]
tobz: To address my comment on this code block specifically, though, what we currently do in `on_new_span` is:

```
let mut labels = Labels::from_record(&Record::new(attrs.values()));

if let Some(parent) = span.parent() {
    if let Some(parent_labels) = parent.extensions().get::<Labels>() {
        labels.extend_from_labels(parent_labels);
    }
}
```

We create our set of labels from the current span's attributes, and then we extend those labels from the parent span's calculated labels, if any. Based on how the code is written in `extend_from_labels`, any overlapping labels in the parent span's labels would overwrite our own labels from the current span.

metrics-tracing-context/src/tracing_integration.rs
tobz: Hmm, interesting. Let me think about this, since I couldn't wrap my head around it on the first read-through. 😅

zohnannor: I added helpful comments so it's easier to digest (hopefully)!

tobz: Yeah, neat.

If you want to try reworking the current code to use this approach, in this same PR, I'd be fine with that... but also happy to review it in a separate PR so we can get this merged sooner.

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -1,28 +1,37 @@
...
 
 impl Labels {
     pub(crate) fn extend_from_labels(&mut self, other: &Labels) {
-        self.0.extend_from_slice(other.as_ref());
+        let new_len = cmp::max(self.as_ref().len(), other.as_ref().len());
+        let additional = new_len - self.as_ref().len();
+        self.0.reserve(additional);
+        for (k, v) in other.as_ref() {
+            self.0.insert(k.clone(), v.clone());
+        }
```
tobz, line 34: Looks like this still needs to be changed to only call `insert` if `k` doesn't exist in `self`.

metrics-tracing-context/tests/integration.rs
```diff
@@ -110,7 +120,200 @@ fn test_basic_functionality() {
...
+            ),
+            None,
+            None,
+            DebugValue::Counter(1),
+        )]
+    );
+}
+
+#[test]
+fn test_record_does_not_overwrite() {
```
tobz, line 292: I think we also need to add a test like this one, but where the overwriting is happening due to a nested parent/child span, showing that the field value in the child span doesn't get overridden by the same field (with a different value) in the parent span.

tobz (changes requested):
Looking good, but it still seems like we don't quite have the label priority logic correct.

metrics-tracing-context/tests/integration.rs
zohnannor: `test_nested_spans` already tests this behavior.

metrics-tracing-context/src/tracing_integration.rs
zohnannor: No, because in that case we would not overwrite the labels via `record`. Added a separate methods for both operations.

metrics-tracing-context/tests/integration.rs
tobz: Ah yeah, weird... not sure how I missed that. 🤔

metrics-tracing-context/src/lib.rs
```diff
@@ -169,34 +175,27 @@ where
...
+                (!span_labels.is_empty()).then(|| {
+                    let (name, labels) = key.clone().into_parts();
+
+                    span_labels.extend(labels.into_iter().map(Label::into_parts));
+
+                    let labels = span_labels
+                        .into_iter()
+                        .map(|(key, value)| Label::new(key, value))
+                        .filter(|label| self.label_filter.should_include_label(&name, label))
+                        .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
zohnannor: Oh! We are filtering only labels coming from `tracing`, got it.

metrics-tracing-context/Cargo.toml
```diff
@@ -32,13 +32,17 @@ itoa = { version = "1", default-features = false }
 metrics = { version = "^0.21", path = "../metrics" }
 metrics-util = { version = "^0.15", path = "../metrics-util" }
 lockfree-object-pool = { version = "0.1.3", default-features = false }
+indexmap = { version = "2.1", default-features = false, features = ["std"] }
 once_cell = { version = "1", default-features = false, features = ["std"] }
 tracing = { version = "0.1.29", default-features = false }
 tracing-core = { version = "0.1.21", default-features = false }
 tracing-subscriber = { version = "0.3.1", default-features = false, features = ["std"] }
+tracing-test = "0.2.4"
```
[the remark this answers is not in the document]
zohnannor: I'll remove it entirely because I added it for debugging (same for other testing-related crates)

zohnannor:
Pretty sad perf stats. Ready for a review though.

```java
     Running benches/layer.rs (/home/zohnannor/dev/metrics/target/release/deps/layer-5bf703a34fc6c1b2)
Gnuplot not found, using plotters backend
layer/base case         time:   [923.19 ps 923.33 ps 923.48 ps]
                        change: [-16.774% -16.758% -16.741%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
layer/no integration    time:   [928.33 ps 928.41 ps 928.52 ps]
                        change: [-14.183% -14.125% -14.074%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
layer/tracing layer only
                        time:   [923.75 ps 923.78 ps 923.81 ps]
                        change: [-14.670% -14.622% -14.579%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe
layer/metrics layer only
                        time:   [25.805 ns 26.988 ns 28.176 ns]
                        change: [+17.291% +21.416% +25.369%] (p = 0.00 < 0.05)
                        Performance has regressed.
layer/full integration  time:   [270.72 ns 270.85 ns 270.98 ns]
                        change: [+50.150% +50.248% +50.344%] (p = 0.00 < 0.05)
                        Performance has regressed.

     Running benches/visit.rs (/home/zohnannor/dev/metrics/target/release/deps/visit-1cf93c53aa04b250)
Gnuplot not found, using plotters backend
visit/record_str        time:   [23.671 ns 23.686 ns 23.699 ns]
                        change: [+137.50% +138.60% +139.73%] (p = 0.00 < 0.05)
                        Performance has regressed.
visit/record_bool[true] time:   [18.172 ns 18.179 ns 18.186 ns]
                        change: [+201.05% +201.53% +201.80%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  4 (4.00%) high mild
visit/record_bool[false]
                        time:   [18.097 ns 18.099 ns 18.101 ns]
                        change: [+199.13% +199.29% +199.44%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  3 (3.00%) high mild
  8 (8.00%) high severe
visit/record_i64        time:   [25.666 ns 25.673 ns 25.681 ns]
                        change: [+92.582% +92.723% +92.862%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
visit/record_u64        time:   [25.482 ns 25.488 ns 25.495 ns]
                        change: [+94.980% +95.075% +95.167%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
visit/record_debug      time:   [163.06 ns 163.16 ns 163.26 ns]
                        change: [+21.900% +22.026% +22.130%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
visit/record_debug[bool]
                        time:   [39.089 ns 39.096 ns 39.104 ns]
                        change: [+73.869% +74.077% +74.278%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
visit/record_debug[i64] time:   [46.002 ns 46.010 ns 46.018 ns]
                        change: [+67.113% +67.556% +67.989%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
visit/record_debug[u64] time:   [43.717 ns 43.731 ns 43.746 ns]
                        change: [+62.503% +62.569% +62.631%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
```

zohnannor:
Cool! Thank you for review! When should we expect the release? 🙂

zohnannor:
@tobz friendly ping~

tobz:
👋🏻

Sorry, life has been very busy 😭 ... although I'm finally on holiday vacation starting today. 😀 Going to try and see if I can get release notes written up today and then cut releases for everything.

tobz:
Released as `metrics-tracing-context@v0.15.0`.

PR #412, opened 2023-11-21
Upgrade hyper to v1

fenollp:
cc https://github.com/metrics-rs/metrics/pull/411
The hyper CHANGELOG https://github.com/hyperium/hyper/blob/master/CHANGELOG.md


tobz:
Looks like this requires unpublished changes to `hyper-tls`. I don't tie anything in `metrics` to unpublished crate versions, so this PR will have to wait until `hyper-tls` cuts a release.

Issue #416, opened 2023-11-29
Can't run clippy


nstinus:
metrics-exporter-prometheus has clippy lints [enabled](https://github.com/metrics-rs/metrics/blob/c37a4077ab67282836911b6a0c65f8fe368ed871/metrics-exporter-prometheus/src/lib.rs#L97). Yet, many are outstanding.

For example:
```
error: use of `or_insert_with` to construct default value
  --> metrics-exporter-prometheus/src/recorder.rs:89:18
   |
89 |                 .or_insert_with(IndexMap::new)
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `or_default()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_or_default
note: the lint level is defined here
  --> metrics-exporter-prometheus/src/lib.rs:97:9
   |
97 | #![deny(clippy::all)]
   |         ^^^^^^^^^^^
   = note: `#[deny(clippy::unwrap_or_default)]` implied by `#[deny(clippy::all)]`

error: could not compile `metrics-exporter-prometheus` (lib) due to 6 previous errors
```

Either:
- downgrade them to warnings or,
- add a check in the ci.

tobz:
I think this should be all set now that #417 has been merged.

PR #417, opened 2023-11-29
Add a clippy ci check (#416)

nstinus:
This change adds a ci test for clippy and quickly addresses the blocking points by downgrading them to warnings if they can't be trivially addressed.


tobz:
Nice, this is good and helpful. 👍🏻

Issue #418, opened 2023-11-29
Make tls support optional and add rustls


nstinus:
Tls support was recently added by forcing the usage of openssl via hyper-tls. It means that either this lib is installed or it needs to be compiled. For users unconcerned with tls, we could remove the dependency. Alternatively, rustls provides a pure rust tls implementation.

bushkov:
I am also facing an issue because of the added dependency on `hyper-tls`.  I have to cross compile a project to a statically linked binary to be able to run on `aarch64-unknown-linux-gnu`, which is currently impossible to do with the dependency on `hyper-tls` added in version 0.13.0 of `metrics-exporter-prometheus`.

Please switch either to `rustls` or provide a way to opt out of HTTPS (which would be my preference since I don't need a secure connection for my use case).

justin13888:
Are there any plans on moving forward with @nstinus' suggestions? On my end, either switching to rustls and removing hyper-tls outright would be acceptable.

tobz:
I left feedback on their PR, which hasn't (yet?) been incorporated. I don't have the time/need to modularize TLS support, so this is waiting on someone to submit a PR doing so, or an update to the aforementioned PR based on my feedback.

tessus:
@tobz if you are ok with switching to rustls, I can create a PR for that. Tomorrow that is. It's way past my bedtime. ;-)

PR #420, opened 2023-11-30
Make tls optional and add rustls support (#418)

nstinus:
This change removed the required dependency to hyper-tls and openssl. The allow tls, clients will now have to enable either the `native-tls` or `rustls-tls` features.

BREAKING: tls isn't enabled by default anymore.

### DIFF
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -20,7 +20,9 @@ keywords = ["metrics", "telemetry", "prometheus"]
 default = ["http-listener", "push-gateway"]
 async-runtime = ["tokio", "hyper"]
 http-listener = ["async-runtime", "hyper/server", "ipnet"]
-push-gateway = ["async-runtime", "hyper/client", "hyper-tls", "tracing"]
+push-gateway = ["async-runtime", "hyper/client", "tracing"]
+native-tls = ["hyper-tls"]
+rustls-tls = ["hyper-rustls"]
 
 [dependencies]
 metrics = { version = "^0.21", path = "../metrics" }
@@ -36,6 +38,7 @@ ipnet = { version = "2", optional = true }
 tokio = { version = "1", features = ["rt", "net", "time"], optional = true }
 tracing = { version = "0.1.26", optional = true }
 hyper-tls = { version = "0.5.0", optional = true }
+hyper-rustls = { version = "0.24.2", optional = true }
 
 [dev-dependencies]
 tracing = "0.1"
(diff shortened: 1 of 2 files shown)


metrics-exporter-prometheus/Cargo.toml
```diff
@@ -20,7 +20,9 @@ keywords = ["metrics", "telemetry", "prometheus"]
 default = ["http-listener", "push-gateway"]
 async-runtime = ["tokio", "hyper"]
 http-listener = ["async-runtime", "hyper/server", "ipnet"]
-push-gateway = ["async-runtime", "hyper/client", "hyper-tls", "tracing"]
+push-gateway = ["async-runtime", "hyper/client", "tracing"]
```
tobz, line 23: When it comes to the core crates (like `metrics`), I generally err on the side of making users opt-in to additional functionality, but that flips the opposite direction for exporters, where users generally want/expect to get the "full fat" support out of the box.

I'd say we should add another feature called `push-gateway-https`, which pulls in `hyper-rustls`, and have the feature be enabled by default.

metrics-exporter-prometheus/Cargo.toml
```diff
@@ -20,7 +20,9 @@ keywords = ["metrics", "telemetry", "prometheus"]
...
+native-tls = ["hyper-tls"]
+rustls-tls = ["hyper-rustls"]
```
tobz, line 25: After doing some reading, it seems like we should probably just switch to using `rustls` entirely.

I don't think the initial implementor choose to use `hyper-tls` specifically for the OS-backed/dynamically-linked aspect, and I don't personally care about esoteric SSL/TLS setups or FIPS compliance or any of that... so I'd rather go with a single solution than force users to have to spend time understanding which one they ought to choose.

tobz (changes requested):

metrics-exporter-prometheus/Cargo.toml
rbtcollins: I hope you won't mind a somewhat random drive-past, but its generally a good practice for mid-layer crates that sit between TLS and applications to not set TLS policy.

The way to make this work comfortably is:
- at a dependency level consume hyper, or whatever other crates you use, with `default-features-false`
- have a feature to turn on the push-gateway *without* specifying the TLS stack to use
- have a different feature to turn on the push-gateway with whatever tls stack you prefer to have by default

Then higher layer crates such as axum-prometheus can depend on your crate, with default-features=false, and the application layer crate using axum-prometheus can specify hyper-tls or hyper-rustls itself.

PR #423, opened 2023-12-12
fix(prom): `RollingSummary` overflow panic

LucioFranco:
This fixes a `RollingSummary` overflow/underflow panic that happens when time goes backwards. In this case, instead of panicking we detect the underflow and will drop the value.

### DIFF
--- a/metrics-exporter-prometheus/src/distribution.rs
+++ b/metrics-exporter-prometheus/src/distribution.rs
@@ -218,15 +218,6 @@ impl RollingSummary {
 
             self.buckets.truncate(self.max_buckets - 1);
             self.buckets.insert(0, Bucket { begin, summary });
-        } else {
-            begin = reftime - self.bucket_duration;
-            while now < begin {
-                begin -= self.bucket_duration;
-            }
-
-            self.buckets.truncate(self.max_buckets - 1);
-            self.buckets.push(Bucket { begin, summary });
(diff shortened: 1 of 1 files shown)


metrics-exporter-prometheus/src/distribution.rs
```diff
@@ -219,7 +219,17 @@ impl RollingSummary {
...
+            // case we will just throw away the value.
+            //
+            // Check `add_value_ts_before_first_bucket` test for reference situation
+            // that has happened already and triggered a panic.
+            if let Some(b) = reftime.checked_sub(self.bucket_duration) {
+                begin = b;
+            } else {
+                return;
+            };
+
```
tobz, line 232: I think we just want to drop this entire `else` block.

Otherwise, we would also have to guard the subtraction on line 234 (`begin -= self.bucket_duration`) but also because if we _really_ want to do this, we would have to deal with wrapping timestamps.

Essentially, if we care enough to try and keep in-the-past samples around and have a bucket for them, then we need to do it right... but if we want to just ignore in-the-past samples due to not wanting to deal with underflow, then we should ignore _all_ in-the-past samples.

tobz (changes requested):

metrics-exporter-prometheus/src/distribution.rs
LucioFranco: done, ive removed the branch and the associated tests.

tobz (approved):

PR #427, opened 2023-12-19
Remove unneeded unsafe from test

rukai:
I replaced unsafe by making the test use a regular variable instead of a constant, any potential performance here is irreverent as its just a test and now we have one less unsafe to audit.

I also added a SAFETY comment for a suspicious looking unsafe.
If you know for sure its fine to have that kind of unsoundness there if its never actually called then I can remove the ambiguity from the phrasing.


rukai:
Thanks!

> I removed the change adding the SAFETY comment to Block<T> because looking at the code, it's not actually safe and I'm going to file an issue to fix that.

Mission accomplished I guess 😎

PR #429, opened 2023-12-19
Try other resolved addresses if the first one fails

vvv:
Problem: `metrics-observer` cannot connect to `raspberrypi.local:8000` -- "error while connecting".

Analysis:

"raspberrypi.local:8000" gets resolved on my machine to
- `[fe80::da3a:ddff:fe3c:8772%15]:8000`
- `192.168.0.100:8000`

The IPv6 `SocketAddr` -- the first item returned by the iterator -- cannot be connected to. The IPv4 address would work, but `metrics-observer` never tries it.

Solution: if the first item, returned by `to_socket_addrs`, cannot be connected to, try the next one. Try them all, if necessary.

### DIFF
--- a/metrics-observer/src/metrics.rs
+++ b/metrics-observer/src/metrics.rs
@@ -114,30 +114,22 @@ impl Runner {
                         let mut state = self.client_state.lock().unwrap();
                         *state = ClientState::Disconnected(None);
                     }
-
-                    // Try to connect to our target and transition into Connected.
-                    let addr = match self.addr.to_socket_addrs() {
-                        Ok(mut addrs) => match addrs.next() {
-                            Some(addr) => addr,
-                            None => {
-                                let mut state = self.client_state.lock().unwrap();
-                                *state = ClientState::Disconnected(Some(
-                                    "failed to resolve specified host".to_string(),
-                                ));
-                                break;
-                            }
-                        },
-                        Err(_) => {
-                            let mut state = self.client_state.lock().unwrap();
-                            *state = ClientState::Disconnected(Some(
-                                "failed to resolve specified host".to_string(),
-                            ));
-                            break;
-                        }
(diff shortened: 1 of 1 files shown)


metrics-observer/src/metrics.rs
```diff
@@ -114,30 +114,22 @@ impl Runner {
...
-                        },
-                        Err(_) => {
-                            let mut state = self.client_state.lock().unwrap();
-                            *state = ClientState::Disconnected(Some(
-                                "failed to resolve specified host".to_string(),
-                            ));
-                            break;
-                        }
+                    // Resolve the target address.
+                    let Ok(mut addrs) = self.addr.to_socket_addrs() else {
```
vvv, line 118: Alternatively, we could limit the results to IPv4 address(es).

```rust
                    let Ok(mut addrs) = self.addr.to_socket_addrs() else {
                        let mut state = self.client_state.lock().unwrap();
                        *state = ClientState::Disconnected(Some(
                            "failed to resolve specified host".to_string(),
                        ));
                        break;
                    };
                    let Some(addr) = addrs.find(|addr| addr.is_ipv4()) else {
                        let mut state = self.client_state.lock().unwrap();
                        *state = ClientState::Disconnected(Some(
                            "failed to resolve specified host as IPv4".to_string(),
                        ));
                        break;
                    };
                    match TcpStream::connect_timeout(&addr, Duration::from_secs(3)) {
                        Ok(stream) => RunnerState::Connected(stream),
                        Err(_) => RunnerState::ErrorBackoff(
                            "error while connecting",
                            Duration::from_secs(3),
                        ),
                    }
```

tobz:
Nice, this is a good improvement. 👍🏻

tobz:
Released as `metrics-observer@v0.3.0`.

PR #435, opened 2024-01-08
Specific endpoint for prometheus exporter health checking

biancamip:
**Problem:**
A service that installs a prometheus metrics exporter will return all metrics data on any request, regardless of the endpoint requested. This presents an issue when doing health checks using a service networking solution such as [Consul](https://www.consul.io/), which performs requests on short intervals and **spreads the response status + body across all server nodes**. Since the **response body contains all metrics collected** and will probably change from one check to the next, the issue is that this behavior will generate a high network traffic, resulting in delays, elevated costs, and so on.

**Solution:** 
For health checking, the response could be as simple as a successful status code with a short body content. I've implemented this for the path "/health", while maintaining the previous behavior for all other cases.

If necessary, this solution could be adjusted to allow configuration of the health path on `PrometheusBuilder` struct initialization, and it could also be disabled in case the current behavior is desired, but I think this minor change would make significant difference for now.


tobz:
@biancamip Apologies on the delay reviewing this. 👋🏻 

I've made two small little changes, but otherwise, I think this is fine to merge as soon as CI passes.

tobz:
Forgot to update the changelog prior to release, but this has no been released as part of `metrics-exporter-prometheus@v0.13.1`.

PR #438, opened 2024-01-15
Update `hashbrown` to `0.14`

fornwall:
As the MSRV of `metrics-util` now is `1.65` we can now update to `hashbrown` version `0.14` (which has MSRV `1.63`).


tobz:
Thanks for your contribution!

tobz:
Released as part of `metrics-util@v0.13.2`.

PR #444, opened 2024-02-07
Rolling summary configuration

cparratto:
Adds bucket duration as configurable element for rolling summaries.

### DIFF
--- a/metrics-exporter-prometheus/src/builder.rs
+++ b/metrics-exporter-prometheus/src/builder.rs
@@ -96,6 +96,7 @@ pub struct PrometheusBuilder {
     #[cfg(feature = "http-listener")]
     allowed_addresses: Option<Vec<IpNet>>,
     quantiles: Vec<Quantile>,
+    bucket_duration: Option<Duration>,
     buckets: Option<Vec<f64>>,
     bucket_overrides: Option<HashMap<Matcher, Vec<f64>>>,
     idle_timeout: Option<Duration>,
@@ -120,6 +121,7 @@ impl PrometheusBuilder {
             #[cfg(feature = "http-listener")]
             allowed_addresses: None,
             quantiles,
+            bucket_duration: None,
             buckets: None,
             bucket_overrides: None,
             idle_timeout: None,
@@ -236,6 +238,22 @@ impl PrometheusBuilder {
         }
 
         self.quantiles = parse_quantiles(quantiles);
+        Ok(self)
+    }
+
+    /// Sets the default bucket duration for rolling summaries
+    ///
+    /// Buckets will be cleared after this interval expires
+    ///
+    /// ## Errors
+    ///
+    /// If `value` less than 1 error will be thrown
+    pub fn set_bucket_duration(mut self, value: Duration) -> Result<Self, BuildError> {
+        if value.is_zero() {
+            return Err(BuildError::ZeroBucketDuration);
+        }
+
+        self.bucket_duration = Some(value);
         Ok(self)
     }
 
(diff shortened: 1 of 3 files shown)


cparratto:
I would like to be able to configure these values. Happy to make any other changes required to get this upstream

tobz:
I'll take a look at this in the next day or two.

tobz (changes requested):
If we're going to expose the configuration of rolling summaries, we need to expose both the number of buckets as well as the bucket width.

cparratto:
> If we're going to expose the configuration of rolling summaries, we need to expose both the number of buckets as well as the bucket width.

Sounds good. I will add this to the PR now.

metrics-exporter-prometheus/src/distribution.rs
```diff
@@ -34,9 +34,16 @@ impl Distribution {
...
-    pub fn new_summary(quantiles: Arc<Vec<Quantile>>) -> Distribution {
-        let summary = RollingSummary::default();
-        Distribution::Summary(summary, quantiles, 0.0)
+    pub fn new_summary(
+        quantiles: Arc<Vec<Quantile>>,
+        bucket_duration: Duration,
+        bucket_count: u32,
+    ) -> Distribution {
+        Distribution::Summary(
+            RollingSummary::new(NonZeroU32::new(bucket_count).unwrap(), bucket_duration),
```
tobz, line 43: We should actually take a `NonZeroU32` for `bucket_count` rather than just unwrapping here.

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -239,6 +243,38 @@ impl PrometheusBuilder {
...
+    /// ## Errors
+    ///
+    /// If `value` less than 1 error will be thrown
+    pub fn set_bucket_count(mut self, count: u32) -> Result<Self, BuildError> {
+        if count == 0 {
+            return Err(BuildError::ZeroBucketCount);
+        }
+
+        self.bucket_count = Some(count);
+        Ok(self)
```
tobz, line 275: In conjunction with my comment a little further down about taking `NonZeroU32` directly in `Distribution::new_summary`, we should also make a corresponding change here... essentially, change `self.bucket_count` to `Option<NonZeroU32>` and then this code can become something like:

```rust
let count = NonZeroU32::new(count).ok_or(BuildError::ZeroBucketCount)?;

self.bucket_count = Some(count);
Ok(self)
```

metrics-exporter-prometheus/src/distribution.rs
```diff
@@ -95,7 +108,10 @@ impl DistributionBuilder {
             return Distribution::new_histogram(buckets);
         }
 
-        Distribution::new_summary(self.quantiles.clone())
+        let b_duration = self.bucket_duration.map_or(Duration::from_secs(20), |d| d);
+        let b_count = self.bucket_count.map_or(3, |c| c);
```
tobz, line 112: Since we're now referencing these default in two places -- here, and `RollingSummary::default` -- we should lift them both to constants at the top of the file... like `DEFAULT_SUMMARY_BUCKET_COUNT`, etc.

(We can't construct the `NonZeroU32` in a const fashion, so just the `u32` itself will be fine for now.)

tobz (changes requested):
Thanks for adding the change to update the bucket count, too.

Left some feedback just to clean things up.

metrics-exporter-prometheus/src/builder.rs
cparratto: I changes the function signature to just take NonZeroU32 as an input param which removes the need for BuildError::ZeroBucketCount. Please let me know if that works for you.

metrics-exporter-prometheus/src/distribution.rs
cparratto: All fixed up!

tobz (approved):
Looks like there's some Clippy lints, but overall, this looks good enough to me to merge. I'll deal with the lints as part of some other changes I need to make in the same area.

tobz:

PR #450, opened 2024-02-17
chore/hyper 1.1 upgrade

evaneaston:
- update to hyper 1.1
  - factor large `PrometheusBuilder`.`build` function into separate functions 
  - move those functions into separate modules to reduce some of the feature driven uses
  - add an http-listener integration test

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/builder.rs
+++ b/metrics-exporter-prometheus/src/exporter/builder.rs
@@ -1,4 +1,3 @@
-
 use std::collections::HashMap;
 #[cfg(feature = "push-gateway")]
 use std::convert::TryFrom;
@@ -186,7 +185,7 @@ impl PrometheusBuilder {
         self.quantiles = parse_quantiles(quantiles);
         Ok(self)
     }
-    
+
     /// Sets the bucket width when using summaries.
     ///
     /// Summaries are rolling, which means that they are divided into buckets of a fixed duration
--- a/metrics-exporter-prometheus/src/exporter/builder.rs
+++ b/metrics-exporter-prometheus/src/exporter/builder.rs
@@ -29,7 +29,9 @@ use crate::recorder::{Inner, PrometheusRecorder};
 use crate::registry::AtomicStorage;
 use crate::{common::BuildError, PrometheusHandle};
 
-use super::{ExporterConfig, ExporterFuture};
+use super::ExporterConfig;
+#[cfg(any(feature = "http-listener", feature = "push-gateway"))]
+use super::ExporterFuture;
 
 /// Builder for creating and installing a Prometheus recorder/exporter.
 pub struct PrometheusBuilder {
--- a/metrics-exporter-prometheus/src/exporter/builder.rs
+++ b/metrics-exporter-prometheus/src/exporter/builder.rs
@@ -1,12 +1,10 @@
+
 use std::collections::HashMap;
 #[cfg(feature = "push-gateway")]
 use std::convert::TryFrom;
-use std::future::Future;
 #[cfg(feature = "http-listener")]
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
-#[cfg(any(feature = "http-listener", feature = "push-gateway"))]
 use std::num::NonZeroU32;
-#[cfg(any(feature = "http-listener", feature = "push-gateway"))]
 use std::sync::RwLock;
 #[cfg(any(feature = "http-listener", feature = "push-gateway"))]
 use std::thread;
@@ -31,41 +29,7 @@ use crate::recorder::{Inner, PrometheusRecorder};
 use crate::registry::AtomicStorage;
 use crate::{common::BuildError, PrometheusHandle};
 
-#[cfg(any(feature = "http-listener", feature = "push-gateway"))]
-pub type ExporterFuture = Pin<Box<dyn Future<Output = Result<(), hyper::Error>> + Send + 'static>>;
-
-#[derive(Clone)]
-enum ExporterConfig {
-    // Run an HTTP listener on the given `listen_address`.
-    #[cfg(feature = "http-listener")]
-    HttpListener { listen_address: SocketAddr },
-
-    // Run a push gateway task sending to the given `endpoint` after `interval` time has elapsed,
-    // infinitely.
-    #[cfg(feature = "push-gateway")]
-    PushGateway {
-        endpoint: Uri,
-        interval: Duration,
-        username: Option<String>,
-        password: Option<String>,
-    },
-
-    #[allow(dead_code)]
-    Unconfigured,
-}
-
-impl ExporterConfig {
-    #[cfg_attr(not(any(feature = "http-listener", feature = "push-gateway")), allow(dead_code))]
-    fn as_type_str(&self) -> &'static str {
-        match self {
-            #[cfg(feature = "http-listener")]
-            Self::HttpListener { .. } => "http-listener",
-            #[cfg(feature = "push-gateway")]
-            Self::PushGateway { .. } => "push-gateway",
-            Self::Unconfigured => "unconfigured,",
-        }
-    }
-}
+use super::{ExporterConfig, ExporterFuture};
 
 /// Builder for creating and installing a Prometheus recorder/exporter.
 pub struct PrometheusBuilder {
@@ -220,7 +184,7 @@ impl PrometheusBuilder {
         self.quantiles = parse_quantiles(quantiles);
         Ok(self)
     }
-
+    
     /// Sets the bucket width when using summaries.
     ///
     /// Summaries are rolling, which means that they are divided into buckets of a fixed duration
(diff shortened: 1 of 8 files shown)


evaneaston:
Sounds like a plan.  That's the direction I was going to go.   Now to find some more time.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Clone)]
+struct Inner {
+    handle: PrometheusHandle,
+    allowed_addresses: Option<Vec<IpNet>>,
+}
+
+impl HttpListeningExporter {
+    async fn serve(&self, listener: std::net::TcpListener) -> Result<(), hyper::Error> {
+        // TODO - address panic possibility
+        let listener = TcpListener::from_std(listener).unwrap();
```
evaneaston, line 32: Something needs to be done about the possibility of panic.  There's no good way to generate a hyper::Error.

One options is for this crate to switch to a crate-specific Error for the exporter future result error.  This would of course be a breaking change.

evaneaston:
@tobz Took a shot and reorganizing the exporters.  I removed the nested closure of the http exporter as some might find it easier to read.  I need to revisit Cargo.toml.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+                Ok((stream, _)) => stream,
+                Err(e) => {
+                    warn!("Error accepting connection. Ignoring request. Error: {:?}", e);
+                    continue;
+                }
+            };
+
+            let remote_addr = match stream.peer_addr() {
+                Ok(remote_address) => remote_address.ip(),
+                Err(e) => {
```
tobz, line 45: The logic here makes sense, but I think what we might want to do is hoist up the logic from `handle_http_request`, for checking if the address is allowed, and then pass down the result of that into `process_stream`, and eventually, `handle_http_request`.

If we basically do something like...

```rust
let is_allowed = self.inner.allowed_addresses
    .map(|addrs| {
        stream.peer_addr().map(|peer_addr| {
            addrs.iter().any(|addr| addr.contains(&peer_addr))
        })
        .unwrap_or_else(|| {
            warn!("...");
            false
        })
    })
    .unwrap_or(true);
```

Then we always get `true` if no allowlist is configured, `false` if an allowlist _is_ configured but we can't detect the peer address, and whatever the result of the `any` call is if both are present.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    async fn process_stream(&self, stream: TcpStream, remote_address: IpAddr) {
+        let inner = self.inner.clone();
+        let service = service_fn(move |req: Request<body::Incoming>| {
+            let inner = inner.clone();
+            async move { Self::handle_http_request(&inner, remote_address, &req) }
+        });
+
+        tokio::task::spawn(async move {
```
tobz, line 62: `tokio::taskl::spawn` -> `tokio::spawn`, just to be a little more idiomatic.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+impl HttpListeningExporter {
+    async fn serve(&self, listener: std::net::TcpListener) -> Result<(), hyper::Error> {
+        // TODO - address panic possibility
+        let listener = TcpListener::from_std(listener).unwrap();
+
+        loop {
+            let stream = match listener.accept().await {
+                Ok((stream, _)) => stream,
+                Err(e) => {
+                    warn!("Error accepting connection. Ignoring request. Error: {:?}", e);
```
tobz, line 38: Add a dedicated `error` field on the event that we assign the error value to:

```rust
warn!(error = ?e, "Error accepting connection. Ignoring request.");
```

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+                    continue;
+                }
+            };
+
+            let remote_addr = match stream.peer_addr() {
+                Ok(remote_address) => remote_address.ip(),
+                Err(e) => {
+                    warn!("Error obtaining remote address. Ignoring request. Error: {:?}", e);
```
tobz, line 46: Add a dedicated `error` field on the event that we assign the error value to.

(see previous comments about logging for format example)

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+        let service = service_fn(move |req: Request<body::Incoming>| {
+            let inner = inner.clone();
+            async move { Self::handle_http_request(&inner, remote_address, &req) }
+        });
+
+        tokio::task::spawn(async move {
+            if let Err(err) =
+                HyperHttpBuilder::new().serve_connection(TokioIo::new(stream), service).await
+            {
+                warn!("Error serving connection.  Error: {:?}", err);
```
tobz, line 66: Add a dedicated `error` field on the event that we assign the error value to.

(see previous comments about logging for format example)

metrics-exporter-prometheus/src/exporter/builder.rs
```diff
@@ -409,117 +352,36 @@ impl PrometheusBuilder {
     #[cfg_attr(docsrs, doc(cfg(any(feature = "http-listener", feature = "push-gateway"))))]
     #[cfg_attr(not(feature = "http-listener"), allow(unused_mut))]
     pub fn build(mut self) -> Result<(PrometheusRecorder, ExporterFuture), BuildError> {
+
```
tobz, line 356: Stray newline.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -0,0 +1,131 @@
...
+    let listener = std::net::TcpListener::bind(listen_address)
+        .and_then(|listener| {
+            listener.set_nonblocking(true)?;
+            Ok(listener)
+        })
+        .map_err(|e| BuildError::FailedToCreateHTTPListener(e.to_string()))?;
+
+    let exporter = HttpListeningExporter { inner: Arc::new(Inner { handle, allowed_addresses }) };
+
+    Ok(Box::pin(async move { exporter.serve(listener).await }))
```
tobz, line 118: This can just be `Ok(Box::pin(exporter.serve(listener)))`, pretty sure.

metrics-exporter-prometheus/src/exporter/push_gateway.rs
```diff
@@ -0,0 +1,136 @@
...
+                        let status = response.status();
+                        let status =
+                            status.canonical_reason().unwrap_or_else(|| status.as_str());
+                        let body = response
+                            .into_body()
+                            .collect()
+                            .await
+                            .map(Collected::aggregate)
+                            .map_err(|_| ())
+                            .map(|mut b| b.copy_to_bytes(b.remaining()))
```
tobz, line 59: Might be a little simpler to use `Collected::to_bytes` above and then you could drop this line entirely, I think.

metrics-exporter-prometheus/tests/http_listener_integration_test.rs
```diff
@@ -0,0 +1,91 @@
...
+        .unwrap_or_else(|e| panic!("Error decoding response body: {:?}", e));
+
+    (status, body_string)
+}
+
+fn find_available_port() -> Option<u16> {
+    (25000_u16..26000).find(|&port| match std::net::TcpListener::bind(("localhost", port)) {
+        Ok(_) => true,
+        Err(_) => false,
+    })
```
tobz: If you bind to port 0 instead, it'll just pick an ephemeral port. You can then call `TcpListener::local_addr` to nab the port.

tobz (changes requested):
Overall, this is looking really good. Some of the `hyper` stuff is unfortunate given how much more verbose it's gotten, but alas. :)

Left a bunch of cleanup comments, and I think once those are addressed, I can take one more pass since it was a lot to digest at first.

metrics-exporter-prometheus/src/exporter/http_listener.rs
evaneaston: I was just trying to preserve the previous semantics of failing with a `BuildError` if the port was unavailable, before the future is even created and awaited in a runtime.  This proposal would be a non-breaking change API-wise, but break consumers relying on a `BuildError` when the port is unavailable. 

But it's entirely your call.

metrics-exporter-prometheus/src/exporter/http_listener.rs
evaneaston: Sure  I was attempting to move the `is_allowed` check entirely into the `handle_http_request` method to offload from blocking the `serve` loop.  I also was trying to separate the three responsibilities:

 - `serve` = listen and loop
 - `process_stream` = spawn task
 - `handle_http_request` = authentication and processing
 
This, of course, necessitated the wrapping `allowed_addresses` withing the `Arc<Inner>`.  But this allowed me to only have to pass one server state var into `handle_http_request` along with the request state.  And this also adds a double `Arc` dereference to get to the recorder.

So a bunch of tradeoffs.  But your call.

metrics-exporter-prometheus/src/exporter/http_listener.rs
evaneaston: 👍 TIL

metrics-exporter-prometheus/tests/http_listener_integration_test.rs
evaneaston: :cool:

metrics-exporter-prometheus/src/exporter/http_listener.rs
tobz: That's still the case, as far as I can see?

`PrometheusBuilder::build` is what is calling `new_http_listener`. It was already going to return the error for binding the port because the way you have it creates the underlying `std::net::TcpListener` there already. Converting it to a `tokio::net::TcpListener` doesn't re-bind or anything, and it should only throw errors in very specific conditions like not being able to register the FD to epoll or whatever.

metrics-exporter-prometheus/src/exporter/http_listener.rs
tobz: That makes sense, but I would say that, logically, checking if the remote address is  allowed should happen at the connection level, not per-request... even though, really, it's only going to be one request per connection in most cases.

I think the tradeoffs are worth it to get the level of specificity closer to optimal.

metrics-exporter-prometheus/src/exporter/http_listener.rs
evaneaston: 👍 Yes.  By changing serve to consume `self` instead of `&self`, the move is implicit.

evaneaston:
@tobz Would you like me to rebase and --signoff on my stuff.  It'll require a diverging force push.

tobz:
> @tobz Would you like me to rebase and --signoff on my stuff. It'll require a diverging force push.

If that's cool with you, that'd be nice. Force pushing is fine.

evaneaston:
@tobz I pushed the tweak to fix the workflow fail.

tobz:
@evaneaston Just noticed the merge to get up-to-date with `main`.

I'll take a final pass shortly and hopefully get this merged today. I think it's close enough that it's good enough to merge, since there's a lot of other changes queued up for release and I'll likely be doing some stylistic cleanup to my own preferences anyways.

tobz (approved):

tobz:

PR #460, opened 2024-03-08
feat: add capability to purge old histogram data

mnpw:
Copy of #451 

---

# What

- add purge_timeout option to PrometheusBuilder
- run a purger that purges based on the purge_timeout

Implements third way as [prescribed here](https://github.com/metrics-rs/metrics/issues/245#issuecomment-974155299) to purge old histogram data:
> - update the builder to generate a future which both drives the Hyper server future as well as a call to get_recent_metrics on an interval

Fixes https://github.com/metrics-rs/metrics/issues/245

### DIFF
--- a/metrics-exporter-prometheus/src/builder.rs
+++ b/metrics-exporter-prometheus/src/builder.rs
@@ -99,6 +99,7 @@ pub struct PrometheusBuilder {
     buckets: Option<Vec<f64>>,
     bucket_overrides: Option<HashMap<Matcher, Vec<f64>>>,
     idle_timeout: Option<Duration>,
+    purge_timeout: Option<Duration>,
     recency_mask: MetricKindMask,
     global_labels: Option<IndexMap<String, String>>,
 }
@@ -123,6 +124,7 @@ impl PrometheusBuilder {
             buckets: None,
             bucket_overrides: None,
             idle_timeout: None,
+            purge_timeout: None,
             recency_mask: MetricKindMask::NONE,
             global_labels: None,
         }
@@ -307,6 +309,17 @@ impl PrometheusBuilder {
         self
     }
 
+    /// Sets the purge timeout for metrics.
+    ///
+    /// If a purge timeout is set, the purger will call `.render()` on the registry, causing
+    /// the values from histograms to be drained out. This ensures that stale histogram values
+    /// do not persist indefinitely.
+    #[must_use]
+    pub fn purge_timeout(mut self, timeout: Option<Duration>) -> Self {
+        self.purge_timeout = timeout;
+        self
+    }
+
     /// Adds a global label to this exporter.
     ///
     /// Global labels are applied to all metrics.  Labels defined on the metric key itself have precedence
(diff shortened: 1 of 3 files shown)


metrics-exporter-prometheus/src/builder.rs
```diff
@@ -465,11 +478,40 @@ impl PrometheusBuilder {
...
+                    .enable_all()
+                    .build()
+                    .map_err(|e| BuildError::FailedToCreateRuntime(e.to_string()))?;
+
+                thread::Builder::new()
+                    .name(thread_name.to_owned())
+                    .spawn(move || runtime.block_on(purger))
+                    .map_err(|e| BuildError::FailedToCreateRuntime(e.to_string()))?;
+            };
+        }
```
tobz, line 513: No need for the conditional logic here: just spawn the future directly.

(We already document that this method must be called from within a Tokio runtime or else it will panic.)

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -465,11 +478,40 @@ impl PrometheusBuilder {
...
         #[cfg(feature = "http-listener")]
         let allowed_addresses = self.allowed_addresses.take();
+        let purge_timeout = self.purge_timeout.take();
 
         let exporter_config = self.exporter_config.clone();
         let recorder = self.build_recorder();
         let handle = recorder.handle();
 
+        // use the handle to recorder
+        // #[cfg(not(feature = "push-gateway"))]
```
tobz, line 488: Delete this line.

metrics-exporter-prometheus/src/recorder.rs
```diff
@@ -273,4 +287,10 @@ impl PrometheusHandle {
     pub fn render(&self) -> String {
         self.inner.render()
     }
+
+    /// Purges registry's histogram data by draining it into the distribution. This should be
+    /// called periodically to prevent the accumulation of histogram samples.
+    pub fn purge(&self) {
```
tobz, line 293: I think we should just change the wording overall from "purge" to "upkeep", and make this `run_upkeep`.

Purging to me sounds like getting rid of, when realistically we're just doing periodic cleanup work.

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -361,6 +363,17 @@ impl PrometheusBuilder {
         self
     }
 
+    /// Sets the purge timeout for metrics.
+    ///
+    /// If a purge timeout is set, the purger will call `.render()` on the registry, causing
+    /// the values from histograms to be drained out. This ensures that stale histogram values
+    /// do not persist indefinitely.
```
tobz, line 370: Similar note here about just changing the wording overall from "purge"/"purging" to "upkeep".

I would also remove the specific bit about calling `.render()`, since it doesn't actually do that anymore. Just be generic, something like:

```rust
/// Sets the upkeep interval.
///
/// The upkeep task handles periodic maintenance operations, such as draining histogram data,
/// to ensure that all recorded data is up-to-date and prevent unbounded memory growth. 
```

metrics-exporter-prometheus/src/builder.rs
```diff
@@ -128,6 +129,7 @@ impl PrometheusBuilder {
             buckets: None,
             bucket_overrides: None,
             idle_timeout: None,
+            purge_timeout: None,
```
tobz, line 132: I think we should actually enable this by default.

Thinking more about it, it's a quality of life improvement to avoid unbounded memory growth for people with a high rate of histogram metrics, or who scrape/push their Prometheus exporter infrequently.

Probably 5 seconds as the default?

tobz (changes requested):
This is looking good, but just a few notes/requests to try and make this a little more generic/understandable to folks.

tobz (approved):
Perfect. Nice and simple. 👍🏻

tobz:

PR #461, opened 2024-03-10
Add retain functions to Registry

kathoum:
This adds public functions retain_counters, retain_gauges, retain_histograms, that work like HashMap::retain

### DIFF
--- a/metrics-util/src/registry/mod.rs
+++ b/metrics-util/src/registry/mod.rs
@@ -410,6 +410,48 @@ where
         let shard_read = shard.read().unwrap_or_else(PoisonError::into_inner);
         shard_read.raw_entry().from_key_hashed_nocheck(hash, key).map(|(_, v)| v.clone())
     }
+
+    /// Retains only counters specified by the predicate.
+    ///
+    /// Remove all counters for which f(&k, &c) returns false. This operation proceeds
+    /// through the "subshards" in the same way as `visit_counters`.
+    pub fn retain_counters<F>(&self, mut f: F)
+    where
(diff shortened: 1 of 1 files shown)


kathoum:
Design-wise, I'm not sure about the choice on the function signature:
`FnMut(&K, &mut S::Counter) -> bool` or `FnMut(&K, &S::Counter) -> bool`

I went with `&mut` because there's already a write lock, so the code has exclusive access to the underlying `S::Counter`, and it shouldn't make much difference in the common case that `Counter = Arc<impl CounterFn>`.

But I understand if we prefer not allowing `mut` access to objects inside the storage.

kathoum:
PR updated, changed the visitors to accept immutable references

metrics-util/src/registry/mod.rs
```diff
@@ -456,6 +498,15 @@ mod tests {
...
 
+        registry.get_or_create_counter(&Key::from_name("baz"), |_| ());
+
+        let mut n = 0;
+        registry.retain_counters(|k, _| {
+            n += 1;
+            k.name().starts_with("foo")
+        });
+        assert_eq!(n, 2);
+
```
tobz, line 509: I think you'd want to copy the assert from lines 512-513 and stick them after adding the `baz` counter, to show that we now have two counters in the registry. After that, do the same thing after the `retain_counters` call to show it goes back down to one counter after it deletes `baz`.

tobz (changes requested):

tobz (approved):
Looks good to me. 👍🏻

tobz:
Released in `metrics-util@v0.16.3`.

PR #466, opened 2024-03-23
feat: get_recent_metrics as method from PrometheusHandle

lpj145:
If we choose to only uses the Recorder, I think with that implementation we could achieve a lot of implementations, like translate to otlp, request a rest api, print to stdout and some other things.

### DIFF
--- a/metrics-exporter-prometheus/src/recorder.rs
+++ b/metrics-exporter-prometheus/src/recorder.rs
@@ -287,6 +287,11 @@ impl PrometheusHandle {
         self.inner.render()
     }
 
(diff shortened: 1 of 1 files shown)


tobz:
Hey, sorry for the delay in getting around to this PR.

I'm going to close this, as this feature is not something I'm looking to currently add. Many people have asked for this before, but the truth is that installing a "Prometheus" exporter and not even using any Prometheus-specific features, just grabbing the metrics, is a weird usage of the crate that I don't want to encourage.

I'm happy to accept a PR for a generic recorder that _only_ supports getting manual snapshots of the registry, though.

PR #477, opened 2024-04-29
Update lib.rs

Kinrany:
There seems to be a typo in the explanation of histograms.

It says "50% of requests are slower than 500ms and 99% of requests are slower than [~2000ms]”. That doesn't make sense because the latter already implies the former.

Flipping from "slower" to faster seems to make more sense statistically, but there may be an even better way to phrase this.


tobz (approved):
Good catch. Thanks for submitting this.

tobz:

Kinrany:
Thank you!

Issue #474, opened 2024-04-24
Make it possible to extract `&'static T` from `Cow<'static, T>`.


Dav1dde:
I am implementing a recorder from `metrics` to `sentry::metrics` and Sentry currently uses `std::cow::Cow<'static, str>` for tags and metric names. If it was possible to extract a `&'static str` from a `SharedStr` many `to_owned` calls would be saved.

I tried implementing it myself but it seems easy to get wrong.

tobz:
The lifetime used does not change the fact the wrapper might be holding an owned value, which means you _cannot_ provide a static reference to it, so it has to be something like `as_borrowed(&self) -> Option<&'a T>`, or an impl to convert to `std::borrow::Cow`, which is also possible.

PR #478, opened 2024-05-01
Implement conversion from cow to std cow

Dav1dde:
Closes: #474

### DIFF
--- a/metrics/src/cow.rs
+++ b/metrics/src/cow.rs
@@ -137,7 +137,7 @@ pub struct Cow<'a, T: Cowable + ?Sized + 'a> {
     _lifetime: PhantomData<&'a T>,
 }
 
-impl<'a, T> Cow<'a, T>
+impl<T> Cow<'_, T>
 where
     T: Cowable + ?Sized,
 {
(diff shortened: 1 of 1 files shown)


Dav1dde:
Did some experiments with `miri` and it seems to be happy.

metrics/src/cow.rs
```diff
@@ -178,6 +178,35 @@ where
...
+    pub fn as_borrowed(&self) -> Option<&'a T> {
+        match self.metadata.kind() {
+            Kind::Owned | Kind::Shared => None,
+            Kind::Borrowed => {
+                // SAFETY: We know the contained data is borrowed from 'a, we're simply
+                // restoring the original immutable reference and returning a copy of it.
+                Some(unsafe { &*T::borrowed_from_parts(self.ptr, &self.metadata) })
+            }
+        }
+    }
```
tobz, line 209: Unless there's a compelling reason to have this be a standalone method, we should just move the logic down into `impl From<Cow<'a, T>> for std::borrow::Cow<'a, T>`.

tobz (changes requested):
The logic is sound, but I think we just need to clean up the ergonomics a little.

metrics/src/cow.rs
Dav1dde: The problem with that is, the `From` impl always requires a `Clone` (or taking ownership, but the way the `SharedString` is exposed it is usually through a reference), the `as_borrowed()` allows for a decision.

Dav1dde:
Removed the `as_borrowed` and only left the cow impl.

tobz:

PR #484, opened 2024-05-10
Relax bounds on some `metrics_util::registry::Registry` methods

david-perez:
I was thinking through https://github.com/metrics-rs/metrics/discussions/481 when I noticed this possible bounds relaxation.

---

The `Registry` struct definition requires `K: Hashable`, which is
superfluous since it's already required in the impl block.

Moreover, some methods don't require the full `K: Clone + Eq + Hashable`
bounds. Some methods only require `K: Eq + Hashable`, others only
require `K: Hashable`, while others don't require any bounds at all on
`K`.

This commit splits the single impl block into three impl blocks, so
users that have keys that don't satisfy all bounds can still make use of
some methods.

### DIFF
--- a/metrics-util/src/registry/mod.rs
+++ b/metrics-util/src/registry/mod.rs
@@ -47,7 +47,6 @@ type RegistryHashMap<K, V> = HashMap<K, V, BuildHasherDefault<RegistryHasher>>;
 /// `Registry` is optimized for reads.
 pub struct Registry<K, S>
 where
-    K: Hashable,
     S: Storage<K>,
 {
     counters: Vec<RwLock<RegistryHashMap<K, S::Counter>>>,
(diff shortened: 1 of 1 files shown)


metrics-util/src/registry/mod.rs
```diff
@@ -47,7 +47,6 @@ type RegistryHashMap<K, V> = HashMap<K, V, BuildHasherDefault<RegistryHasher>>;
 /// `Registry` is optimized for reads.
 pub struct Registry<K, S>
 where
-    K: Hashable,
     S: Storage<K>,
 {
     counters: Vec<RwLock<RegistryHashMap<K, S::Counter>>>,
```
david-perez, line 52: Out of curiosity, why are we explicitly depending on a `hashbrown::HashMap` for this type? Isn't the standard library's identical?

tobz: `std::collections::HashMap<K, V>` is based on `hashbrown` now, yes, but we use the raw entry API, which is still not yet stabilized in the standard library.

tobz (approved):
Looks good to me. 👍🏻

tobz:
Released in `metrics-util@v0.17.0`.

PR #489, opened 2024-06-01
refactor: use hyper-rustls

tessus:
From https://github.com/metrics-rs/metrics/pull/420#discussion_r1435328304:

> After doing some reading, it seems like we should probably just switch to using `rustls` entirely.
> 
> I don't think the initial implementor choose to use `hyper-tls` specifically for the OS-backed/dynamically-linked aspect, and I don't personally care about esoteric SSL/TLS setups or FIPS compliance or any of that... so I'd rather go with a single solution than force users to have to spend time understanding which one they ought to choose.

This PR switches to `hyper-rustls`.

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/push_gateway.rs
+++ b/metrics-exporter-prometheus/src/exporter/push_gateway.rs
@@ -3,7 +3,6 @@ use std::time::Duration;
 use http_body_util::{BodyExt, Collected, Full};
 use hyper::body::Bytes;
 use hyper::{header::HeaderValue, Method, Request, Uri};
-use hyper_tls::HttpsConnector;
 use hyper_util::{client::legacy::Client, rt::TokioExecutor};
 use tracing::error;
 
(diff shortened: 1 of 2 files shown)


tessus:
@tobz what do you think?

metrics-exporter-prometheus/src/exporter/push_gateway.rs
```diff
@@ -19,7 +18,12 @@ pub(super) fn new_push_gateway(
     handle: PrometheusHandle,
 ) -> ExporterFuture {
     Box::pin(async move {
-        let https = HttpsConnector::new();
+        let https = hyper_rustls::HttpsConnectorBuilder::new()
+            .with_native_roots()
+            .expect("no native root CA certificates found")
+            .https_only()
```
tobz, line 24: This change will only allow pushing to HTTPS endpoints, whereas the old code allows for either HTTP or HTTPS.

This should instead of `https_or_http`.

tobz (changes requested):
Apologies for the delay in reviewing.

Just one small change that needs to be made and then I think it should be good to go.

metrics-exporter-prometheus/src/exporter/push_gateway.rs
tessus: yes, of course. sorry about that.

tessus: done

tobz (approved):

tobz:
I'm going to try to get this released today.

tobz:
Released in `metrics-exporter-prometheus@v0.15.1`.

tessus:
Thank you for merging and for the release! Have a great evening.

PR #496, opened 2024-06-27
Set plain/text Content-Type for prometheus exporter

snaggen:
Playing around with different tools to view metrics, I found that some tool complained about not having Content-Type text/plain. So this PR fixes that.

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/http_listener.rs
+++ b/metrics-exporter-prometheus/src/exporter/http_listener.rs
@@ -3,6 +3,7 @@ use std::net::SocketAddr;
 use http_body_util::Full;
 use hyper::{
     body::{self, Bytes, Incoming},
+    header::{HeaderValue, CONTENT_TYPE},
     server::conn::http1::Builder as HyperHttpBuilder,
     service::service_fn,
     Request, Response, StatusCode,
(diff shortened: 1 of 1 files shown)


metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -69,10 +70,12 @@ impl HttpListeningExporter {
         req: &Request<Incoming>,
     ) -> Response<Full<Bytes>> {
         if is_allowed {
-            Response::new(match req.uri().path() {
+            let mut response = Response::new(match req.uri().path() {
                 "/health" => "OK".into(),
                 _ => handle.render().into(),
-            })
+            });
+            response.headers_mut().append(CONTENT_TYPE, HeaderValue::from_static("text/plain"));
```
sgrebnov, line 77: Should this be `text/plain; version=0.0.4`?
https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

sgrebnov:
Encountered with the same issue when using this crate and Datadog

sgrebnov:
I've manually applied this fix and confirmed that the Prometheus exporter is now compliant with the DataDog Prometheus scraper.

tobz:
Seems reasonable to me! 👍🏻

tobz (approved):

tobz:
Released as `metrics-exporter-prometheus@v0.15.3`.

PR #498, opened 2024-07-08
HTTP exporter can listen on unix domain socket (UDS) as well as TCP

scottopell:
HTTP over Unix Domain Sockets is useful in situations where the thing exposing telemetry does not have networking configured, but does have easy shared disk.

In my use case, this occurs because we run a container without external networking, but still want to collect telemetry from the container. This is trivial with a Unix Domain Socket mounted in a volume share.


metrics-exporter-prometheus/src/exporter/builder.rs
```diff
@@ -133,6 +138,26 @@ impl PrometheusBuilder {
...
     }
 
+    /// Configures the exporter to expose an HTTP Unix Domain Socket listener that functions as a [scrape endpoint].
+    ///
+    /// The HTTP listener that is spawned will respond to GET requests on any request path.
+    ///
+    /// Running in HTTP listener mode is mutually exclusive with the push gateway i.e. enabling the
+    /// HTTP listener will disable the push gateway, and vise versa.
+    ///
+    /// Defaults to disabled, if enabled, listens on the specified path
```
tobz, line 148:
```suggestion
    /// Configures the exporter to expose an HTTP listener that functions as a [scrape endpoint], listening on a Unix Domain socket at the given path.
    ///
    /// The HTTP listener that is spawned will respond to GET requests on any request path.
    ///
    /// Running in HTTP listener mode is mutually exclusive with the push gateway i.e. enabling the
    /// HTTP listener will disable the push gateway, and vise versa.
    ///
    /// Defaults to disabled.
```

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -29,31 +58,52 @@ impl HttpListeningExporter {
...
+            }
+        });
+    }
+
+    fn check_tcp_allowed(&self, stream: &TcpStream) -> bool {
+        if let Some(addrs) = &self.allowed_addresses {
+            if let Ok(peer_addr) = stream.peer_addr() {
+                return addrs.iter().any(|addr| addr.contains(&peer_addr.ip()));
+            }
+        }
```
tobz, line 87: This causes a regression by allowing the request through even if we failed to get the peer address.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -29,31 +58,52 @@ impl HttpListeningExporter {
...
+        true
+    }
+
+    #[cfg(feature = "uds-listener")]
+    async fn serve_uds(&self, listener: &UnixListener) -> Result<(), std::io::Error> {
+        loop {
+            let (stream, _) = listener.accept().await?;
```
tobz, line 94: I think we'd want this to match how we do it in the TCP listener, where we log a warning but still keep accepting connections.

metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -9,18 +9,47 @@ use hyper::{
...
+use std::path::PathBuf;
 use tokio::net::{TcpListener, TcpStream};
+#[cfg(feature = "uds-listener")]
+use tokio::net::{UnixListener, UnixStream};
 use tracing::warn;
 
 use crate::{common::BuildError, ExporterFuture, PrometheusHandle};
 
-struct HttpListeningExporter {
+pub struct HttpListeningExporter {
```
tobz, line 21: Why is this `pub` now?

tobz (changes requested):

tobz:
Thanks for your contribution!

tobz:
Released as `metrics-exporter-prometheus@v0.15.3`.

PR #500, opened 2024-07-31
Replace num_cpus::get with std::thread::available_parallelism

rukai:
[std::thread::available_parallelism](https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html) was stabilized in rust 1.59, so there should be no problem in using it in this crate with an MSRV of 1.70

To remain consistent with the existing usage of `num_cpus::get` we default to a value of 1 if info on parallelism is inaccessible for whatever reason.
The old implementation did a `std::cmp::max(1, ..)` but that was not needed with num_cpus since it was documented to return a max of 1, and its not needed now since we handle an error by using 1.


tobz (approved):
Seems fine to me! 👍🏻

tobz:
Released as part of `metrics-util@v0.18.0`.

PR #504, opened 2024-08-19
Derive Debug trait for all structs and enums

joshka:
This allows all the structs and enums to easily be placed in a struct
which derives Debug itself.

The metrics::recorder::Recorder and metrics_tracing_context::LabelFilter
traits now also extend Debug.

### DIFF
--- a/metrics-tracing-context/src/label_filter.rs
+++ b/metrics-tracing-context/src/label_filter.rs
@@ -1,6 +1,6 @@
 //! Label filtering.
 
-use std::collections::HashSet;
+use std::{collections::HashSet, fmt::Debug};
 
 use metrics::{KeyName, Label};
 
--- a/metrics-tracing-context/src/label_filter.rs
+++ b/metrics-tracing-context/src/label_filter.rs
@@ -1,6 +1,6 @@
 //! Label filtering.
 
-use std::{collections::HashSet, fmt::Debug};
+use std::collections::HashSet;
 
 use metrics::{KeyName, Label};
 
--- a/metrics-util/src/layers/fanout.rs
+++ b/metrics-util/src/layers/fanout.rs
@@ -105,7 +105,9 @@ pub struct Fanout {
 
 impl fmt::Debug for Fanout {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Fanout").finish_non_exhaustive()
+        f.debug_struct("Fanout")
+            .field("recorders_len", &self.recorders.len())
+            .finish_non_exhaustive()
     }
 }
 
@@ -166,7 +168,9 @@ pub struct FanoutBuilder {
 
 impl fmt::Debug for FanoutBuilder {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("FanoutBuilder").finish_non_exhaustive()
+        f.debug_struct("FanoutBuilder")
+            .field("recorders_len", &self.recorders.len())
+            .finish_non_exhaustive()
     }
 }
 
--- a/metrics-util/src/layers/fanout.rs
+++ b/metrics-util/src/layers/fanout.rs
@@ -1,10 +1,11 @@
-use std::sync::Arc;
+use std::{fmt, sync::Arc};
 
 use metrics::{
     Counter, CounterFn, Gauge, GaugeFn, Histogram, HistogramFn, Key, KeyName, Metadata, Recorder,
     SharedString, Unit,
 };
 
+#[derive(Debug)]
 struct FanoutCounter {
     counters: Vec<Counter>,
 }
@@ -35,6 +36,7 @@ impl From<FanoutCounter> for Counter {
     }
 }
 
+#[derive(Debug)]
 struct FanoutGauge {
     gauges: Vec<Gauge>,
 }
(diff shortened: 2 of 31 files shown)


tobz:
This PR is unreasonably wide reaching. What is it exactly that you're trying to accomplish?

joshka:
The core motivation was to within a library (Ratatui) to be able to centrally define, describe and store `Counter`/`Gauge`/`Histogram` away from the line of code where changes are recorded, to store them for usage from that code. However these don't implement Debug, which means that I'd have to replace all the derived Debug implementations with manual ones.

After implementing those, I decided that it probably made sense to implement it for everything (per https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits) (I'm assuming this was a mistake now and I should have checked). Several of the items where deriving Debug were problematic, and I went with what seemed to be the right approach of manual implementations / adding a Debug constraint to the traits.

Are there any particular parts of this PR that fall into the unreasonable bucket more than others, or is your perspective more that you're unconvinced of the usefulness of generally implement Debug in this crate? My perspective is that implementing Debug is generally a useful thing to do on any item that would potentially ever need to be stored in another item. Is that something you might have differences of opinion on?

tobz:
Overall, I think just blanket deriving `Debug` has the potential to rapidly clutter debug output, especially with some of the larger structs that are being updated here... but that's not a blocking opinion.

My two no-nos here:

1. deriving `Debug` on types in binary crates (`metrics-benchmark`, `metrics-observer`)
2. requiring `Debug` on traits

For #1, this mostly just strikes me as clutter. If there was a strong need to have debug output, they'd have `Debug` derived or implemented already.

For #2, it doesn't make sense to make traits a supertrait of common traits _unless_ they actually require that trait's functionality for their own behavior (i.e. the supertrait's effective behavior is the sum of the subtraits) or they would otherwise solve a large ergonomics issue, like having `Send` and `Sync` as subtraits on a trait with async functions, where it makes it cleaner downstream by not forcing callers to have to add ` + Send + Sync` to all of their where clauses.

In this case, neither of those scenarios is at play, and more tactically, this would only matter if you're holding a `Box<dyn Recorder>` (or other smart pointer), which... I don't see why anybody's code would be doing that. :P

So, all of that said: if we remove the derives from the binary crates, and from traits, I'd be OK with merging the rest.

joshka:
Makes sense. I've removed the derives from the metrics-observer and benches, and removed the trait changes.
Rebased on main to pick up the other merged changes.

Clippy lints are unrelated to this change (will submit another PR to clean these up).

metrics-tracing-context/src/label_filter.rs
```diff
@@ -1,6 +1,6 @@
 //! Label filtering.
 
-use std::collections::HashSet;
+use std::{collections::HashSet, fmt::Debug};
```
tobz, line 3: Missed a spot here.

metrics-util/src/layers/fanout.rs
```diff
@@ -100,6 +103,12 @@ pub struct Fanout {
     recorders: Vec<Box<dyn Recorder>>,
 }
 
+impl fmt::Debug for Fanout {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Fanout").finish_non_exhaustive()
```
tobz, line 108: TIOLI: could always add a field like `recorders_len` with `self.recorders.len()` just to have _some_ meaningful information.

(Same below with `FanoutBuilder`.)

tobz (commented):
Just one little missed spot and one nit.

metrics-util/src/layers/fanout.rs
joshka: Copilot suggested that too, but I removed it as I thought it seemed a little off. I was only 60/40 on removing it though, so given that you've also suggested it I'm inclined to put it back in.

joshka: Done in c16f43b

metrics-tracing-context/src/label_filter.rs
joshka: Done in c16f43b

joshka:
Rebased on main.
Fixed the unnecessary import.
Added _len debug fields on Fanout, FanoutBuilder, Router, and RouterBuilder.

joshka:
No probs.
For more on how this is useful to me, see https://github.com/ratatui/ratatui/pull/1351, in particular the code where I centralize the initialization of metrics for each struct that uses them (e.g. https://github.com/ratatui/ratatui/pull/1351/files#diff-2c068b77d24467c86dd276124445d84fa48752b9c38e5f444a54f7a27ea2e0e7). I think this pattern makes it easy to use consistent IDs and language in the descriptions.

```rust
static METRICS: Lazy<Metrics> = Lazy::new(Metrics::new);

#[derive(Debug)]
struct Metrics {
    pub clear_region_count: Counter,
    pub clear_region_duration: Histogram,
    pub draw_count: Counter,
    pub draw_duration: Histogram,
    pub append_lines_count: Counter,
    pub append_lines_duration: Histogram,
    pub hide_cursor_duration: Histogram,
    pub show_cursor_duration: Histogram,
    pub get_cursor_position_duration: Histogram,
    pub set_cursor_position_duration: Histogram,
    pub size_duration: Histogram,
    pub window_size_duration: Histogram,
    pub flush_count: Counter,
    pub flush_duration: Histogram,
}
```

tobz:
Released as part of `metrics@v0.24.0`/`metrics-util@v0.18.0`/`metrics-tracing-context@v0.17.0`/`metrics-exporter-prometheus@v0.16.0`/`metrics-exporter-tcp@v0.11.0`.

PR #505, opened 2024-08-19
Use ratatui instead of tui-rs for the terminal UI

joshka:
Ratatui is a maintained fork of tui-rs. See https://ratatui.rs/ for more
information.

Changes the backend to use crossterm instead of termion. This makes it
possible to use the terminal UI on Windows.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.70.0"
+channel = "1.74.0"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,5 @@
 [toolchain]
+# Note that this is greater than the MSRV of the workspace (1.70) due to metrics-observer needing
+# 1.74, while all the other crates only require 1.70. See
+# https://github.com/metrics-rs/metrics/pull/505#discussion_r1724092556 for more information.
 channel = "1.74.0"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,8 +44,14 @@ jobs:
     name: Test ${{ matrix.rust_version }}
     runs-on: ubuntu-latest
     strategy:
+      # 1.70 is the MSRV for the project, which currently does not match the version specified in
+      # the rust-toolchain.toml file as metrics-observer requires 1.74 to build. See
+      # https://github.com/metrics-rs/metrics/pull/505#discussion_r1724092556 for more information.
       matrix:
-        rust_version: ['stable', 'nightly']
+        rust_version: ['stable', 'nightly', '1.70']
+        include:
+          - rust_version: '1.70'
+            exclude-packages: '--exclude metrics-observer'
     steps:
     - uses: actions/checkout@v3
     - name: Install Protobuf Compiler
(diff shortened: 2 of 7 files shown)


rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.70.0"
+channel = "1.74.0"
```
tobz, line 2: Nope, we're not bumping the project-wide MSRV just for a single binary crate.

tobz (changes requested):
In general, this seems fine?

I left one comment about a portion that needs to be changed, however.

rust-toolchain.toml
joshka: The MSRV of the other crates in the workspace doesn't change here - they can still be built with 1.70.0, but the version of the rust toolchain used to build the entire workspace does unfortunately need to change, or somehow just this this crate needs to be built with a later version. Putting a rust-toolchain.toml config in the metrics-observer works when building at that folder level, but building the workspace still uses the root rust version.

This was introduced due to the following problem:
```shell
❯ cargo build            
error: package `metrics-observer v0.4.0 (/Users/joshka/local/metrics/metrics-observer)` cannot be built because it requires rustc 1.74.0 or newer, while the currently active rustc version is 1.70.0
```
The same error occurs when building the workspace with `cargo build --workspace`

We could remove the metrics-observer from the workspace if that would help, but I think that might be sub optimal. Is there another way to get around this?

We could also use Ratatui 0.25.0, the last version which supported Rust 1.70. The delta is about 400 commits - which 30% of the total commits of the entire repo.

Incidentally Rust 1.74 is in almost the same place as Rust 1.70 was metrics last did a MSRV bump for metrics. It will be 7 versions behind current when 1.81 is released in 2 weeks time. Do you have any downstream crates that are constrained this far back?

tobz: > The MSRV of the other crates in the workspace doesn't change here - they can still be built with 1.70.0, but the version of the rust toolchain used to build the entire workspace does unfortunately need to change, or somehow just this this crate needs to be built with a later version.

Fair point, and my mistake.

> We could remove the metrics-observer from the workspace if that would help, but I think that might be sub optimal.

I agree it would be suboptimal, although still a reasonable approach.

> We could also use Ratatui 0.25.0, the last version which supported Rust 1.70. The delta is about 400 commits - which 30% of the total commits of the entire repo.

This seems _more_ suboptimal: switching to another replacement crate and not even getting on the latest version.

> Do you have any downstream crates that are constrained this far back?

No, but maintaining a lower MSRV is just good practice (IMO) for foundational crates.

After reading through and thinking about it as I was writing all of this, I think leaving this change in place is _fine_, but we _should_ update the `test-matrixed` CI step ([here](https://github.com/metrics-rs/metrics/blob/main/.github/workflows/ci.yml#L48)) to add the MSRV to the matrix of versions to test against. This will provide a counterbalance to `rust-toolchain.toml` no longer being pinned to the MSRV.

joshka: Makes sense. I added a test using rust 1.70, with a parameter to exclude testing the metrics-observer package.
I also added some docs showing why this is there (linking back to your comment here).

joshka:
> Looks like this needs a base branch update, but I think those Clippy errors are a factor of using 1.74 now... so we can ignore them and I'll clean them up once this is merged.

Sounds good - I merged main into this branch and fixed the clippy lint in the metrics-observer, but left the other ones.

tobz (approved):

PR #507, opened 2024-08-20
Add blanket implementations of `Recorder` for some pointer-like types

### LINKED ISSUE #506: Routing more than one pattern to the same `Recorder`
[`metrics_util::layers::RouterBuilder::add_route`](https://docs.rs/metrics-util/latest/metrics_util/layers/struct.RouterBuilder.html#method.add_route) takes ownership of the recorder, which means it can only handle one pattern/mask combination.

Could we make it take in an `Iterator` of patterns, or an `Arc<R: Recorder>`?

david-perez:
`Rc`, `Arc`, and `Box`.

Closes #506.

### DIFF
--- a/metrics/src/recorder/mod.rs
+++ b/metrics/src/recorder/mod.rs
@@ -248,6 +248,7 @@ mod tests {
         // This test simply ensures that if a boxed recorder is handed to us to install, and another
         // recorder has already been installed, that we drop th new boxed recorder instead of
         // leaking it.
+        #[derive(Debug)]
         struct TrackOnDropRecorder(Arc<AtomicBool>);
 
         impl TrackOnDropRecorder {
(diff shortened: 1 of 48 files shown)


metrics/src/recorder/mod.rs
```diff
@@ -57,6 +57,95 @@ pub trait Recorder {
...
 }
 
+// Blanket implementations.
+
+impl<T> Recorder for Rc<T>
+where
+    T: Recorder,
+{
+    fn describe_counter(&self, key: KeyName, unit: Option<Unit>, description: SharedString) {
+        Deref::deref(self).describe_counter(key, unit, description)
```
tobz, line 67: You should be able to just do `(**self)`, or even `self.as_ref()`, which I think are both a bit more idiomatic than `Deref::deref(self)`. Dealer's choice on which one you choose... but just not `Deref::deref(self)`. 😂

tobz (changes requested):
Just one small nit to have the code be a little more idiomatic.

tobz:
Closing in favor of #507. (explanation is in the new PR)

PR #510, opened 2024-08-25
Fix clippy lints

joshka:


metrics-exporter-prometheus/src/exporter/http_listener.rs
```diff
@@ -59,11 +59,11 @@ impl HttpListeningExporter {
                     continue;
                 }
             };
-            self.process_tcp_stream(stream).await;
+            self.process_tcp_stream(stream);
         }
     }
 
-    async fn process_tcp_stream(&self, stream: TcpStream) {
+    fn process_tcp_stream(&self, stream: TcpStream) {
```
joshka, line 66: This change is because the method doesn't actually await at all. I'm assuming that this is the right change, based on the usual case that a loop that handles incoming TCP requests usually wants to spawn a task and continue waiting for the next incoming requests. The other thing that could be meaningful here would be to wait for the request to be handled (e.g. await the task::spawn call). I think it's more likely the first, but could be wrong.

metrics-exporter-prometheus/tests/http_listener_integration_test.rs
```diff
@@ -36,7 +36,7 @@ mod http_listener_test {
             let labels = vec![Label::new("wutang", "forever")];
             let key = Key::from_parts("basic_gauge", labels);
             let gauge = recorder.register_gauge(&key, &METADATA);
-            gauge.set(-3.14);
+            gauge.set(-1.23);
```
joshka, line 39: Clippy complained about this being too close to PI - easier to change it than to add an allow. Oddly, this was the only line it complained about (there are other similar places where this is used).

metrics-util/src/registry/recency.rs
```diff
@@ -217,6 +217,7 @@ impl GenerationalAtomicStorage {
 /// tracking recency does not matter, despite their otherwise tight coupling.
 pub struct Recency<K> {
     mask: MetricKindMask,
+    #[allow(clippy::type_complexity)]
```
joshka, line 220: The same recommendation to extract a type alias here. This code was easier to understand, but I couldn't think of a good name for the extracted type(s)

metrics/benches/macros.rs
```diff
@@ -38,19 +38,19 @@ fn macro_benchmark(c: &mut Criterion) {
         })
     });
     group.bench_function("global_initialized/no_labels", |b| {
-        let _ = metrics::set_global_recorder(TestRecorder::default());
+        let _ = metrics::set_global_recorder(TestRecorder);
```
joshka, line 41: Calling default on a unit struct is unnecessary.

tobz (commented):
Left a bunch of comments and nits.

metrics-exporter-prometheus/src/exporter/http_listener.rs
joshka: I'm not sure if you're saying you want to allow the lint here, or that the change is fine.

metrics-exporter-prometheus/src/formatting.rs
```diff
@@ -321,7 +321,7 @@ mod tests {
             let as_chars = result.chars().collect::<Vec<_>>();
 
             if let Some(c) = as_chars.first() {
-                assert_eq!(false, invalid_metric_name_start_character(*c),
+                assert!(!invalid_metric_name_start_character(*c),
```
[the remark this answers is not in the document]
joshka: Done in b7cb2ad. Using iterators simplifies this code a bit too.

metrics-util/src/layers/router.rs
```diff
@@ -195,7 +195,7 @@ mod tests {
         let mut builder = RouterBuilder::from_recorder(MockTestRecorder::new());
         builder
             .add_route(MetricKindMask::COUNTER, "foo", MockTestRecorder::new())
-            .add_route(MetricKindMask::GAUGE, "bar".to_owned(), MockTestRecorder::new())
+            .add_route(MetricKindMask::GAUGE, "bar", MockTestRecorder::new())
```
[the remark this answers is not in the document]
joshka: Done in 6eb34fd

metrics-util/src/registry/recency.rs
joshka: I'm going to leave this one as just being an allow

tobz:
Released as part of `metrics@v0.24.0`/`metrics-util@v0.18.0`/`metrics-tracing-context@v0.17.0`/`metrics-exporter-prometheus@v0.16.0`.

PR #512, opened 2024-09-05
Add blanket implementations of `Recorder` for some pointer-like types

### LINKED ISSUE #506: Routing more than one pattern to the same `Recorder`

tobz:
(I messed up the remote branch in #507 and it was just easier to fix it by making a new branch and PR.)

`Rc`, `Arc`, and `Box`.

Closes #506.

### DIFF
--- a/metrics-util/src/layers/router.rs
+++ b/metrics-util/src/layers/router.rs
@@ -166,7 +166,7 @@ mod tests {
         predicate::{always, eq},
         Sequence,
     };
-    use std::{borrow::Cow, rc::Rc};
+    use std::{borrow::Cow, sync::Arc};
 
     use super::RouterBuilder;
     use crate::MetricKindMask;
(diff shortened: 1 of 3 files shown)


Nekrolm:
Thanks for implementing this!

Can you also add `+ ?Sized` bound for blanket impl to support `Box<dyn Recorder>` and similar?

Also I can suggest to use macro to not repeat boilerplate
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021
```Rust
macro_rules! impl_recorder {
    (<$t:tt> for $ptr:ty) => {
        impl <$t> Recorder for $ptr
        where $t : Recorder + ?Sized 
        {
            fn describe_counter(&self, key: &str) {
                $t::describe_counter(self, key)
            }
        }
    }
}

impl_recorder!(<T> for &T);
impl_recorder!(<T> for &mut T);
impl_recorder!(<T> for Box<T>);
impl_recorder!(<T> for std::sync::Arc<T>);
impl_recorder!(<T> for std::rc::Rc<T>);
```

tobz:
Yeah, that's a good idea about reducing the boilerplate with a macro, and also adding blanket implementations for borrowed variants.

It also reminded me that we can't have a blanket implementation for `Rc<T>` now since `Recorder` has to be `Sync`.

tobz:
Released as part of `metrics@v0.24.0`/`metrics-util@v0.18.0`.

Issue #511, opened 2024-08-30
`set_global_recorder` should apply `Sync` bound


Nekrolm:
Hi

This is not an issue for any sane implementations, but currently the lack of Sync restrictions makes the `set_global_recorder` unsound: it's possible to create data race in safe code

example:

```Rust
use metrics::Recorder;
use std::cell::Cell;

struct NonSyncRecorder {
    call_counter: Cell<usize>
}

impl NonSyncRecorder {
    fn inc(&self) {
        let val = self.call_counter.take();
        self.call_counter.set(val + 1);
    }
}

impl Recorder for NonSyncRecorder {
    fn describe_counter(&self, _: metrics::KeyName, _: Option<metrics::Unit>, _: metrics::SharedString) {
        self.inc();
    }

    fn describe_gauge(&self, _: metrics::KeyName, _: Option<metrics::Unit>, _: metrics::SharedString) {
        self.inc();
    }

    fn describe_histogram(&self, _: metrics::KeyName, _: Option<metrics::Unit>, _: metrics::SharedString) {
        self.inc();
    }

    fn register_counter(&self, _: &metrics::Key, _: &metrics::Metadata<'_>) -> metrics::Counter {
        todo!()
    }

    fn register_gauge(&self, _: &metrics::Key, _: &metrics::Metadata<'_>) -> metrics::Gauge {
        todo!()
    }

    fn register_histogram(&self, _: &metrics::Key, _: &metrics::Metadata<'_>) -> metrics::Histogram {
        todo!()
    }
}

fn main() {
    metrics::set_global_recorder(NonSyncRecorder { call_counter: Default::default() }).expect("first initilization");
    std::thread::scope(|s| {
        for _ in 0..100 {
            s.spawn(|| {
                for _ in 0..10000 {
                    metrics::describe_counter!("counter", "dummy");
                }
            });
        }
    })
}

```

Miri output (rustc 1.80):

```
 cargo +nightly miri run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.00s
     Running `/home/dmis/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri runner target/miri/x86_64-unknown-linux-gnu/debug/metrics_not_sync_bug`
error: Undefined Behavior: Data race detected between (1) non-atomic write on thread `unnamed-1` and (2) retag write of type `usize` on thread `unnamed-2` at alloc1158. (2) just happened here
[...]

PR #513, opened 2024-09-06
Add Sync bound for set_global_recorder

Nekrolm:
global recorder is shared between threads and must be sync

closes https://github.com/metrics-rs/metrics/issues/511

### DIFF
--- a/metrics-util/src/recoverable.rs
+++ b/metrics-util/src/recoverable.rs
@@ -55,7 +55,7 @@ pub struct RecoverableRecorder<R> {
     handle: Arc<R>,
 }
 
-impl<R: Recorder + 'static> RecoverableRecorder<R> {
+impl<R: Recorder + Sync + Send + 'static> RecoverableRecorder<R> {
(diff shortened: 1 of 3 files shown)


Nekrolm:
I spend some time thinking, if the Send bound is also needed. But seems not:

https://github.com/metrics-rs/metrics/blob/main/metrics/src/recorder/cell.rs#L44

global recorder is leaked via Box::leak and its destructor won't be called

metrics-util/src/recoverable.rs
```diff
@@ -55,7 +55,7 @@ pub struct RecoverableRecorder<R> {
     handle: Arc<R>,
 }
 
-impl<R: Recorder + 'static> RecoverableRecorder<R> {
+impl<R: Recorder + Sync + Send + 'static> RecoverableRecorder<R> {
```
Nekrolm, line 58: Send enforced here by `Arc`

tobz (approved):

tobz:
Released as part of `metrics@v0.24.0`/`metrics-util@v0.18.0`.

PR #514, opened 2024-09-13
add function to get unrendered snapshot of prometheus metrics

rs-sac:
Our software wants to examine some of these metrics without having to parse the rendered output.

### DIFF
--- a/metrics-exporter-prometheus/src/recorder.rs
+++ b/metrics-exporter-prometheus/src/recorder.rs
@@ -25,7 +25,10 @@ pub(crate) struct Inner {
 }
 
 impl Inner {
-    fn get_recent_metrics(&self) -> Snapshot {
+    fn get_recent_metrics<F>(&self, filter: Option<F>) -> Snapshot
+    where
+        F: Fn(&str) -> bool,
+    {
         let mut counters = HashMap::new();
         let counter_handles = self.registry.get_counter_handles();
         for (key, counter) in counter_handles {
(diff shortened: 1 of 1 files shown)


tobz:
👋🏻 

I appreciate the contribution, but I'm going to close this PR as this feature is not something I'm looking to add. Others have asked for this before, including another PR (#466), and I don't want to encourage this sort of "everything but the kitchen sink" approach to the Prometheus exporter.

As mentioned in the above PR, I'm willing to accept a PR for a generic recorder that provides direct access to the registry, or a snapshot, as this would fit into the happy path of using [`metrics_util::layers`](https://docs.rs/metrics-util/latest/metrics_util/layers/index.html) to customize the global recorder.

tobz:
Exactly.

You'd write this new hypothetical recorder implementation, and then use [`FanoutBuilder`](https://docs.rs/metrics-util/latest/metrics_util/layers/struct.FanoutBuilder.html) to send to both this new recorder _and_ the Prometheus recorder, and then install the resulting `Fanout` as your global recorder.

rs-sac:
I think I get what you're saying, and it's different from what I originally thought.  But what do you think about my suggestion of eavesdropping via a Stack?

With a Fanout, you have to store everything twice, which struck me as suboptimal.  If a recorder in a Stack eavesdrops on registrations and maintains its own lookup table, you only store the underlying metrics once.

tobz:
Apologies... for some reason, I thought I had already responded back to your last comment. 😅 

You could definitely do _something_ like you propose, where a given layer simply creates wrapped versions of the handles returned from the layer below. You would still be storing everything twice, though, because the actual _calls_ to update a metric -- increment counter, set gauge, record histogram -- never go through `Recorder` but through the handles (`Counter`, etc) themselves.

This means you need to return your own handle, where the implementation captures the value and stores it, and then forwards it to the underlying handle... hence the storing it twice part.

In a few issues, the concept has come up of abstracting the storage an exporter uses to allow for a more building blocks-style of creating an exporter, where perhaps there could be some natural interception point: you could imagine using something like `Arc<MyMetricsStorage>`, where the actual exporter gets a copy and you keep your own copy for interrogation. This can be difficult because some exporters end up mutating the metrics storage: consuming values from `AtomicBucket`, removing metrics when they expire, and so on.

This is why fanning out strikes me as the option you actually want: you end up with as little or as much complexity as you want in terms of how they're stored, when they're cleared, and so on.

PR #523, opened 2024-10-01
Introduce support for thread-local default recorder

### LINKED ISSUE #502: Support thread-local default recorder
`with_local_recorder` requires the logic to be inside a closure, which isn't always feasible. Imagine for example running tests: cargo compiles a single binary per target, so if you have multiple tests producing metrics, they will all share the same global recorder. It isn't possible to use a closure unless you combine all your tests into a single one.

The `tracing` crate provides a solution to this in the form of [`set_default`](https://docs.rs/tracing/latest/tracing/subscriber/fn.set_default.html), which allows you to install a thread-local default runtime. This works for `cargo test` as each test runs in a separate thread.

mateiidavid:
Currently, the only way to install a thread-local recorder is to use the provided function `with_local_recorder` function. In asynchronous contexts, especially where single-threaded runtimes are used, using locally scoped recorders can be cumbersome since `with_local_recorder` does not support async execution (unless the closure itself spawns a runtime).

In order to provide an API that overcomes this limitation, for users that want to make use of locally-scoped default recorders, we make it so that the guard can be returned through a newly introduced `set_local_default` function that takes in a trait object. The returned guard is the same type we are using for `with_local_recorder` and when dropped will reset the thread local variable.

The change does not introduce any breaking changes to the existing API. It does however make minimal changes to the internals in order to uphold the safety guarantees that were previously in place. Since the guard is no longer tied to a scope that also encloses the reference, we need to have an explicit lifetime on the guard to guarantee the recorder is not dropped before the guard is. We achieve this through a `PhantomData` type which shouldn't result in any runtime overhead.

Closes #502

---

### Review notes

Thanks a lot in advance for the review! :) Just to preface this, it's not often that I have to reason with unsafe code, I'm happy to make any necessary changes around the code and poke into the internals some more if needed. Sorry if any of this is hard to grok.

* **Motivation**: I find that #502 outlines the motivation for the change. I saw that a tracing-style `set_default()` was a welcome addition so I thought I'd go ahead and implement the change.

* **Safety**: Based on the `SAFETY` comments, my understanding is that we guarantee the pointer we store in thread-local storage is going to be valid since it does not outlive the reference to the `Recorder` trait object by virtue of being enclosed in the function's scope. Naturally, if we return a guard, we can no longer make the same assumptions. Although the compiler will not complain, we're susceptible to footguns if users move or drop their recorders.

I found a lifetime to be easiest to use here to guarantee memory safety, I'm sure there are other alternatives, but I wanted to ensure we do not make any breaking changes, and that we keep the change as minimal as possible.

Unfortunately, I had no inspiration for an automated test, but I did test it manually. In `metrics/src/recorder/mod.rs`, I added the following test

```diff
+    #[test]
+    fn it_compiles() {
+        let local = NoopRecorder;
+        let _guard = crate::set_local_default(&local);
+        drop(local);
+    }
+
```

Without the changes, it successfully compiles and runs. Based on my understanding of the code, this could potentially cause undefined behaviour at runtime since we can hold a reference to a value in memory that has essentially been invalidated.

```sh
:; cargo test recorder::tests::it_compiles -- --nocapture
running 1 test
test recorder::tests::it_compiles ... ok
```

With the changes, we'll get a compiler error:

```sh
:; cargo test recorder::tests::it_compiles -- --nocapture
   Compiling metrics v0.23.0 (/Users/matei.david/workspace/prophecy/metrics/metrics)
error[E0505]: cannot move out of `local` because it is borrowed
   --> metrics/src/recorder/mod.rs:328:14
    |
326 |         let local = NoopRecorder;
    |             ----- binding `local` declared here
327 |         let _guard = crate::set_local_default(&local);
    |                                               ------ borrow of `local` occurs here
328 |         drop(local);
    |              ^^^^^ move out of `local` occurs here
329 |     }
    |     - borrow might be used here, when `_guard` is dropped and runs the `Drop` code for type `LocalRecorderGuard`
```

* **Tests**: I wanted to still have _some_ confidence that the change will not break anything and that it will work as we expect it to. To this end, I added a simple test that exercises a recorder that increments a counter. We start three different recorders and assert the values are expected at the end (i.e. we don't update the wrong thing and we don't get any segfaults).

I also moved the existing recorder impl. (tracking drops) and the new recorder impl. into a module at the bottom of the test module. I wanted to declutter the tests. Let me know if it doesn't look right, I can always revert it. Happy to do what makes most sense for people actively maintaining this.

### DIFF
--- a/metrics/src/recorder/mod.rs
+++ b/metrics/src/recorder/mod.rs
@@ -1,4 +1,4 @@
-use std::{cell::Cell, ptr::NonNull};
+use std::{cell::Cell, marker::PhantomData, ptr::NonNull};
 
 mod cell;
 use self::cell::RecorderOnceCell;
@@ -130,30 +130,39 @@ impl_recorder!(T, std::sync::Arc<T>);
 /// (thread-local storage) so that it can be accessed by the macros. This guard ensures that the
 /// pointer we store to the reference is cleared when the guard is dropped, so that it can't be used
 /// after the closure has finished, even if the closure panics and unwinds the stack.
-struct LocalRecorderGuard;
+///
+/// ## Note
+///
+/// The guard has a lifetime parameter `'a` that is bounded using a
+/// `PhantomData` type. This upholds the guard's contravariance, it must live
+/// _at most as long_ as the recorder it takes a reference to. The bounded
+/// lifetime prevents accidental use-after-free errors when using a guard
+/// directly through [`crate::set_default_local_recorder`].
+pub struct LocalRecorderGuard<'a> {
+    prev_recorder: Option<NonNull<dyn Recorder>>,
+    phantom: PhantomData<&'a dyn Recorder>,
+}
 
-impl LocalRecorderGuard {
+impl<'a> LocalRecorderGuard<'a> {
     /// Creates a new `LocalRecorderGuard` and sets the thread-local recorder.
-    fn new(recorder: &dyn Recorder) -> Self {
+    fn new(recorder: &'a dyn Recorder) -> Self {
         // SAFETY: While we take a lifetime-less pointer to the given reference, the reference we
-        // derive _from_ the pointer is never given a lifetime that exceeds the lifetime of the
-        // input reference.
+        // derive _from_ the pointer is given the same lifetime of the reference
+        // used to construct the guard -- captured in the guard type itself --
+        // and so derived references never outlive the source reference.
         let recorder_ptr = unsafe { NonNull::new_unchecked(recorder as *const _ as *mut _) };
 
-        LOCAL_RECORDER.with(|local_recorder| {
-            local_recorder.set(Some(recorder_ptr));
-        });
+        let prev_recorder =
+            LOCAL_RECORDER.with(|local_recorder| local_recorder.replace(Some(recorder_ptr)));
 
-        Self
+        Self { prev_recorder, phantom: PhantomData }
     }
 }
 
-impl Drop for LocalRecorderGuard {
+impl<'a> Drop for LocalRecorderGuard<'a> {
     fn drop(&mut self) {
         // Clear the thread-local recorder.
-        LOCAL_RECORDER.with(|local_recorder| {
-            local_recorder.set(None);
-        });
+        LOCAL_RECORDER.with(|local_recorder| local_recorder.replace(self.prev_recorder.take()));
     }
 }
 
@@ -173,6 +182,32 @@ where
     R: Recorder + Sync + 'static,
 {
     GLOBAL_RECORDER.set(recorder)
+}
+
+/// Sets the recorder as the default for the current thread for the duration of
+/// the lifetime of the returned [`LocalRecorderGuard`].
+///
+/// This function is suitable for capturing metrics in asynchronous code, in particular
+/// when using a single-threaded runtime. Any metrics registered prior to the returned
+/// guard will remain attached to the recorder that was present at the time of registration,
+/// and so this cannot be used to intercept existing metrics.
+///
+/// Additionally, local recorders can be used in a nested fashion. When setting a new
+/// default local recorder, the previous default local recorder will be captured if one
+/// was set, and will be restored when the returned guard drops.
+/// the lifetime of the returned [`LocalRecorderGuard`].
+///
+/// Any metrics recorded before a guard is returned will be completely ignored.
+/// Metrics implementations should provide an initialization method that
+/// installs the recorder internally.
+///
+/// The function is suitable for capturing metrics in asynchronous code that
+/// uses a single threaded runtime.
+///
+/// If a global recorder is set, it will be restored once the guard is dropped.
+#[must_use]
+pub fn set_default_local_recorder(recorder: &dyn Recorder) -> LocalRecorderGuard {
+    LocalRecorderGuard::new(recorder)
 }
 
 /// Runs the closure with the given recorder set as the global recorder for the duration.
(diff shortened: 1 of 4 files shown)


tobz (changes requested):

metrics/src/recorder/mod.rs
```diff
@@ -175,6 +188,22 @@ where
...
+/// Any metrics recorded before a guard is returned will be completely ignored.
+/// Metrics implementations should provide an initialization method that
+/// installs the recorder internally.
+///
+/// The function is suitable for capturing metrics in asynchronous code that
+/// uses a single threaded runtime.
+///
+/// If a global recorder is set, it will be restored once the guard is dropped.
+#[must_use]
+pub fn set_local_default(recorder: &dyn Recorder) -> LocalRecorderGuard {
```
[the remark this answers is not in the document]
mateiidavid: Fixed! :)

mateiidavid:
@tobz thank you so much for the initial review.

> Overall, I think the approach/intent is reasonable, but reusing the existing LocalRecorderGuard machinery introduces a problem: we would be clearing out any previously-set local records. This is true for any ordering of sequential calls to set_local_default and with_local_recorder.

 I think that's a great idea. I want to preface by saying that I have only just started poking around tracing's internals so my understanding might be a little bit off in certain places.

I have just pushed a change that allows us to restore the previously set local recorder. I have also added a test in to exercise this. It would've been great to do some type assertions using `Any` + downcasting but that was a bit too involved and I couldn't get it to work.

There is also a bit of an issue with the current implementation, for example, suppose you have something like this:

```rust
    #[test]
    fn it_will_segfault() {
        // Create a recorder and install it as the default local.
        let root_recorder = test_recorders::SimpleCounterRecorder::new();
        let root_guard = crate::set_default_local_recorder(&root_recorder);

        // Install a new recorder.
        let next_recorder = test_recorders::SimpleCounterRecorder::new();
        let _next_guard = crate::set_default_local_recorder(&next_recorder);
    
        // Courtesy of the lifetime, we must drop the guard first.
        drop(root_guard);
        // drop() will now restore the previous recorder (in this case, `None`)
        drop(root_recorder);
  
        // drop() will now restore the previous recorder (in this case `root_recorder`)
        drop(_next_guard);
       // segfaults since root_recorder has been dropped a while back
        crate::counter!("test_counter").increment(20);
    }
```

I haven't really found a great way to work around this. In the `tracing` crate, they move the subscriber when installing it locally. This allows them to actually `Arc` it. I couldn't manage to reproduce the issue we are having here. I'm a little bit lost on how to proceed, here are some ideas that I had:

1. Come up with a type where we can specify a lifetime such that a guard cannot outlive the previously set recorder. This didn't work and was a bit awkward, we also cannot have arbitrary lifetimes stored in thread_local afaict, we'd have to give the raw pointer a static lifetime.
2. Use Arc / RefCell, but this unfortunately would break backwards compatibility since `with_local_recorder` takes a trait object. We could have different guard types and handle this differently, for what it's worth.

e.g.
```rust
fn set_default_local_recorder(recorder: impl Recorder) {}
```
3. Provide documentation. Results in a bit of a leaky abstraction and potential footguns.

I realise some of this may sound a bit off, so please let me know if I got anything wrong. Happy to provide more snippets or links to some of the tracing code I've been looking at. 

Also, as a side note, my hunch for taking a raw pointer here is that we want to preserve ownership rules for the recorder? i.e. allow people to have mutable borrows (which would be impossible if we held on to a borrow ourselves). Not sure it's really important here, but I was curious.

metrics/src/recorder/mod.rs
```diff
@@ -130,30 +130,43 @@ impl_recorder!(T, std::sync::Arc<T>);
...
-    fn new(recorder: &dyn Recorder) -> Self {
+    fn new(recorder: &'a dyn Recorder) -> Self {
         // SAFETY: While we take a lifetime-less pointer to the given reference, the reference we
         // derive _from_ the pointer is never given a lifetime that exceeds the lifetime of the
         // input reference.
+        //
+        // NOTE: the lifetime on the argument passed into the constructor is not
+        // strictly needed. It does protect against accidentally passing in a
+        // reference to a reference to a recorder which would result in invalid
+        // memory being accessed.
```
tobz, line 156: I think this comment is duplicative and _potentially_ untrue.. or at least just very confusing. At face value, we obviously need a lifetime on the guard to tie to the lifetime of the reference for which the guard is protecting.

Were you trying to say something else here?

mateiidavid: I did have some trouble with phrasing this, I'll admit. What I meant is that we can theoretically remove the lifetime position from the recorder reference, e.g.

```rust
fn new(recorder: &dyn Recorder)
```

The compiler elides the lifetime for us. However, that would enable us to pass in a reference to the trait object when creating the guard.

```rust
pub fn set_default_local_recorder(recorder: &dyn Recorder) -> LocalRecorderGuard {
    // We take a reference to the trait obj
   // we will now get a segfault if we try to use the recorder
    LocalRecorderGuard::new(&recorder)
}
```

If we keep the lifetime position in the constructor, then the compiler will complain if we have an incorrect argument passed to the guard constructor. Since this is a bit subtle, I was hoping a well placed comment could help people out in the future. Communicating this succinctly is not easy though.

I can remove the comment or try to reformulate it if you'd like.

tobz: Yeah, I would say remove the `NOTE` comment you added, and alter the existing comment to read:

```
// SAFETY: While we take a lifetime-less pointer to the given reference, the reference we
// derive _from_ the pointer is given the same lifetime of the reference used to construct
// the guard itself, and so derived references never outlive the source reference.
```

metrics/src/recorder/mod.rs
```diff
@@ -175,6 +188,22 @@ where
     GLOBAL_RECORDER.set(recorder)
 }
 
+/// Sets the recorder as the default for the current thread for the duration of
```
tobz, line 191: I think we ought to reword this like so:

```suggestion
/// Sets the recorder as the default for the current thread for the duration of
/// the lifetime of the returned [`LocalRecorderGuard`].
///
/// This function is suitable for capturing metrics in asynchronous code, in particular
/// when using a single-threaded runtime. Any metrics registered prior to the returned
/// guard will remain attached to the recorder that was present at the time of registration,
/// and so this cannot be used to intercept existing metrics.
///
/// Additionally, local recorders can be used in a nested fashion. When setting a new
/// default local recorder, the previous default local recorder will be captured if one
/// was set, and will be restored when the returned guard drops.
```

tobz: I realize that you cribbed this comment from what we have for `set_global_recorder`, although I'm going to end up changing that to match what I proposed here, because I think the original comment is a bit stale now with local recorder support in general.

tobz (changes requested):
Just a few small changes I'd like to see to polish the documentation.

metrics/src/recorder/mod.rs
mateiidavid: Thanks for the suggestion. That reads great!

tobz (approved):

mateiidavid:
thanks for the review! quick update: merging was blocked since all commits had to be signed. I squashed all commits and force pushed thinking it'd help; it didn't.

tobz:
Heh, no worries. :D

I'll merge this once tests pass again.

PR #524, opened 2024-10-02
Fix mixed up documentation for Level::* constants

AMDmi3:


tobz (approved):
Oh wow, I can't believe I missed this during code review when this code was first introduced. 🤦🏻 

Thanks for catching this!

tobz:

PR #527, opened 2024-10-03
Update home dependency

### LINKED ISSUE #526: `home` dependency pin causes compatibility issues with other crates
Currently the [home](https://crates.io/crates/home) crate is pinned to [=0.5.5](https://github.com/metrics-rs/metrics/blob/c06d68e1c72b711559cbb19d028d2dce773e7d65/metrics-exporter-tcp/Cargo.toml#L30) across the metrics crates, this causes issues currently for any project which has a transitive dependency on the very popular [which](https://crates.io/crates/which) crate, which currently targets the latest `0.5.9` release of `home` that came out ~1 year ago.

cgbur:
Updates the dependency so it is more compatible with latest packages (e.g. `which` crate). 

fixes #526

### DIFF
--- a/metrics-exporter-tcp/Cargo.toml
+++ b/metrics-exporter-tcp/Cargo.toml
@@ -27,7 +27,7 @@ tracing = { version = "0.1", default-features = false, features = ["attributes"]
 
 [build-dependencies]
 prost-build = "0.12"
-home = "=0.5.5"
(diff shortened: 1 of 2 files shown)


tobz:
Released as part of `metrics-exporter-tcp@v0.11.0`.

PR #531, opened 2024-10-12
enhancement: add new `record_many` method for histograms

### LINKED ISSUE #509: Add  `record_many` to histogram
I want to reexport tokio metrics, they already provide bucketed data.
To reexport it, you should write such code:

```rs
 let hist = metrics::histogram!("tokio_poll_count_time");
  for (idx, value) in histogram.iter().enumerate() {
      let bucket_value = LOG_BUCKETS[idx];
      for _ in 0..*value {
          hist.record(bucket_value);
      }
   }
```                

Is it possible to add 
```rs
fn record_many(&self, value: f64, count: u64);
````
 to `HistogramFn` to not burn cpu cycles? Or maybe there is a better way?

tobz:
## Context

We've added a new method for recording to histograms: `record_many`. This is exposed on both `Histogram` and `HistogramFn`, with a default implementation provided on `HistogramFn` to make it backwards compatible.

It allows recording a single value multiple times, which can be a common use case when translating from bucketed histograms to histograms in `metrics`, as bucketed histograms may simply provide a single value for a bucket and the count of that value... so representing such a bucket in `metrics` requires recording that value `count` times.

Closes #509.

### DIFF
--- a/metrics/CHANGELOG.md
+++ b/metrics/CHANGELOG.md
@@ -11,7 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `Debug` derive to numerous types. ([#504](https://github.com/metrics-rs/metrics/pull/504))
-- Blanket implementations of `Recorder` over smart pointer representations (i.e. `Arc<T> where T: Recorder`). ([#512](https://github.com/metrics-rs/metrics/pull/512))
+- Blanket implementations of `Recorder` over smart pointer representations (i.e. `Arc<T> where T: Recorder`).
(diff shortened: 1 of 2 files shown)


0xdeafbeef:
But it still inefficiently burns CPU cycles:(

tobz:
Yep, there's still work to do in order to actually have storage implementations provide an optimized implementation for it.

If you're interested in contributing a PR to add that support for `AtomicBucket`, I'm happy to try and guide you. The data structure isn't really designed to allow concurrently claiming multiple write slots in a bucket, but I do _think_ it could be done.

tobz:

PR #532, opened 2024-10-12
enhancement: expose `with_recorder` in the docs

### LINKED ISSUE #473: Make with_recorder public
Can `with_recorder()` be made public?

As discussed on Discord, macros can be hard to use in some cases, and some developers strongly prefer plain functions. It would be desirable to have a public API that allows creating and describing metrics without them. As is, it seems that making `with_recorder()` public is the shortest route to achieve this.

The use case where this came up are wrappers, which wrap an action together with it's metric. While it is possible to do that right now, the types and traits involved are hard to find.

tobz:
## Context

This PR exposes `with_recorder` in the documentation. It was always accessible, since it needs to be in order to be used by the macros... but some users expressed a desire to want it to be more visible because they preferred interacting with it directly instead of using the macros.

Closes #473.


tobz:

PR #535, opened 2024-10-18
Add units to prometheus metric lines

### LINKED ISSUE #426: metrics-exporter-prometheus should use units set by describe macro.
## Feature request 1
The `describe_*!` macros set a unit for each metric but metrics-exporter-prometheus does not currently make use of it.
The prometheus protocol itself does not provide a way to set a unit however [it is documented](https://prometheus.io/docs/practices/naming/) that the correct way to handle units in prometheus is to postfix them to the end of the metric name.

So I propose that given a metric like:
```rust
describe_counter!("connections", Unit::Count, "The total count of opened connections")
register_counter!("connections"),
```
It will export as:
```prometheus
# HELP connections The total count of opened connections
# TYPE connections counter
connections_count{} 1
```

This will allow this library to be truly generic across different metrics protocols, as otherwise the user will need to duplicate unit information in other protocols in order to properly use prometheus.

This feature could be disabled in the prometheus builder if there are cases where the user only uses prometheus and needs greater control.

## Feature request 2

Additionally that same page mentions that applications should prefix metrics with the name of the application.
So maybe we should also have an optional setting in the prometheus builder to set the application name such that the previous export will now export as:
```
# HELP connections The total count of opened connections
# TYPE connections counter
application_connections_count{} 1
```

This follows the same reasoning as the previous feature, other protocols may not include the application name at the beginning of each name resulting in redundant information.

fayalalebrun:
This adds units to the prometheus exporter metric lines as the prometheus naming conventions specify:
https://prometheus.io/docs/practices/naming/

The units without a prometheus equivalent such as kilobytes or bytes_per_second were left as is in order to not lose information given by the user.

I also think `_total` should be appended to all counters. However, that is for a future pull request.

Closes #426.

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/builder.rs
+++ b/metrics-exporter-prometheus/src/exporter/builder.rs
@@ -48,6 +48,7 @@ pub struct PrometheusBuilder {
     upkeep_timeout: Duration,
     recency_mask: MetricKindMask,
     global_labels: Option<IndexMap<String, String>>,
+    enable_unit_suffix: bool,
 }
 
 impl PrometheusBuilder {
@@ -80,6 +81,7 @@ impl PrometheusBuilder {
             upkeep_timeout,
             recency_mask: MetricKindMask::NONE,
             global_labels: None,
+            enable_unit_suffix: false,
         }
     }
 
@@ -281,6 +283,19 @@ impl PrometheusBuilder {
 
         self.buckets = Some(values.to_vec());
         Ok(self)
+    }
+
+    /// Sets whether a unit suffix is appended to metric names.
+    ///
+    /// When this is enabled and the [`Unit`][metrics::Unit] of metric is
+    /// given, then the exported metric name will be appended to according to
+    /// the [Prometheus Best Practices](https://prometheus.io/docs/practices/naming/).
+    ///
+    /// Defaults to false.
+    #[must_use]
+    pub fn set_enable_unit_suffix(mut self, enabled: bool) -> Self {
+        self.enable_unit_suffix = enabled;
+        self
     }
 
     /// Sets the bucket for a specific pattern.
@@ -513,6 +528,7 @@ impl PrometheusBuilder {
             ),
             descriptions: RwLock::new(HashMap::new()),
             global_labels: self.global_labels.unwrap_or_default(),
+            enable_unit_suffix: self.enable_unit_suffix,
         };
 
         PrometheusRecorder::from(inner)
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -1,7 +1,7 @@
 //! Helpers for rendering metrics in the Prometheus exposition format.
 
 use indexmap::IndexMap;
-use metrics::Key;
+use metrics::{Key, Unit};
 
 /// Breaks a key into the name and label components, with optional default labels.
 ///
@@ -64,6 +64,7 @@ pub fn write_metric_line<T, T2>(
     labels: &[String],
     additional_label: Option<(&'static str, T)>,
     value: T2,
+    unit: Option<Unit>,
 ) where
     T: std::fmt::Display,
     T2: std::fmt::Display,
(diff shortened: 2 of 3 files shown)


tobz (commented):
So at a high-level, I'm not opposed to this although it will definitely need to live behind a flag to enable it, with it disabled by default.

Separately, though, I'm curious for your thoughts around how to close the gap between what this PR does and what the ideal semantic suffixing strategy is, as defined by the "Metric and label naming" guide.

For example, if we wanted to have a counter for accumulating time spent (e.g., `process_cpu_seconds_total`), how could we accomplish that with units? Should counters always be getting suffixed with `_total` no matter what?

Another thing is base units. The guide recommends avoiding using non-base units (i.e., prefer seconds over milliseconds/nanoseconds/etc) but in this PR, we're just tacking on whatever the defined unit is. Should we actually be adjusting the value to normalize it to seconds? And do a similar thing for bytes, etc.

fayalalebrun:
I think a flag is a good idea. But just to confirm, you do mean a runtime flag right?

I think it would be a good idea in general to normalize the units, though it will require some refactoring. Let me know if you would like me to add it here. What would make this process easier is if `metrics` stored the unit and the magnitude separately.

One incongruity here is the fact that counters can only take an integer value. For instance, this makes it such that counters in seconds do not make much sense. However, for example prometheus uses f64, sentry uses f64 and otlp allows for the use of either f64 or i64. Perhaps then this should change in `metrics` itself, as the limitation to integers does not match most metrics protocols. See: 
https://docs.rs/sentry-core/0.32.3/sentry_core/metrics/type.CounterValue.html
https://docs.rs/opentelemetry-proto/latest/opentelemetry_proto/tonic/metrics/v1/struct.Sum.html

Another concern are the per second metrics. The advice from prometheus is to let the prometheus implementation compute the rate by itself. However,  the prometheus guidance specifies the "power" unit, which is aggregated over time. So normalizing these but otherwise leaving them as-is could be an acceptable solution.

tobz:
I have a note to circle back around to this to re-review it and think about what we've discussed... time has just been hard to come by. 😅

tobz:
Coming back to this again...

> I think a flag is a good idea. But just to confirm, you do mean a runtime flag right?

Yes, something that would end up being configured by `PrometheusBuilder`.
 
> I think it would be a good idea in general to normalize the units, though it will require some refactoring. Let me know if you would like me to add it here. What would make this process easier is if `metrics` stored the unit and the magnitude separately.

Definitely something for later.

> One incongruity here is the fact that counters can only take an integer value. For instance, this makes it such that counters in seconds do not make much sense. However, for example prometheus uses f64, sentry uses f64 and otlp allows for the use of either f64 or i64. Perhaps then this should change in `metrics` itself, as the limitation to integers does not match most metrics protocols. See: https://docs.rs/sentry-core/0.32.3/sentry_core/metrics/type.CounterValue.html https://docs.rs/opentelemetry-proto/latest/opentelemetry_proto/tonic/metrics/v1/struct.Sum.html

I do realize that not using `f64` makes this more difficult. I've yet to see a good explanation or example given where counting negative values makes sense, since allowing negative and positive values to be mixed ends up giving you something a lot more like a gauge. This is probably the most opinionated part of `metrics`: counters are monotonic, and we enforce it by making users deal with `u64`.

It's very unlikely that I will ever change this, at least before metrics v2.

> Another concern are the per second metrics. The advice from prometheus is to let the prometheus implementation compute the rate by itself. However, the prometheus guidance specifies the "power" unit, which is aggregated over time. So normalizing these but otherwise leaving them as-is could be an acceptable solution.

The per-second units are a hold over.. I don't actually remember why I initially added them, but they're sort of pointless at this stage. Users are typically able to compute rates at query time, so dealing with it client-side is unnecessary, and if you actually only have access to a pre-computed rate (like pulling in metrics from a dependent library to be exporter with your app metrics), then it's probably just easier to manually suffix them as such... but that's a tangent.

I'd say we could just pretend the per-second units don't exist because it's more likely than not that I'll remove them in the future.

tobz (approved):

tobz:
Released as `metrics-exporter-prometheus@v0.16.1`. Apologies for the delay on getting around to this... been a hell of a holiday season. 😅 

Thanks for your contribution!

PR #537, opened 2024-10-30
prometheus: update docs to highlight requirements around upkeep

### LINKED ISSUE #467: Histogram unbounded memory even with run_upkeep
Thanks for a wonderful library. After the recent merge of https://github.com/metrics-rs/metrics/pull/460 I tested the upkeep and idle timeout configs but still see unbounded memory usage if /metrics endpoint is not being called.
Is this still expected even after upkeep?

Testing methodology:
- Using axum with axum-prometheus
- Using /metrics
- Using `wrk` to benchmark an API with a simple health endpoint
- Memory keep on increasing unbounded
- Code has idle_timeout of 60s and upkeep of 5s
- If I now open /metrics endpoint, memory eventually reduces and the more I open it periodically, the more stable the memory is

So, is my observed behavior still expected even after the introduction of upkeep?

tobz:
## Context

As stated in the PR title, this change updates the docs for `metrics-exporter-prometheus` to highlight that in certain scenarios, the caller becomes responsible for ensuring the upkeep is run, which helps prevent unbounded memory growth in scenarios where the scrape endpoint is hit infrequently.

Closes #467.

### DIFF
--- a/metrics-exporter-prometheus/CHANGELOG.md
+++ b/metrics-exporter-prometheus/CHANGELOG.md
@@ -8,6 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- Updated the crate-level documentation, and the documentation for [`PrometheusBuilder::build_recorder`] and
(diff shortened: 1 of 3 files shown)


tobz:
Released as `metrics-exporter-prometheus@v0.16.1`.

Issue #533, opened 2024-10-14
`Fanout` and some other utilities in `metrics-util` is not `Sync`


ChieloNewctle:
`metrics::set_global_recorder` requires `Sync` now from the version `0.18.0`.
However, `metrics-util` still uses `Box<dyn Recorder>` for some utilities like `Fanout`, which is not `Sync`.

Maybe some new `Sync` types can be introduced for global recorders?

Thanks.

tobz:
Closed via #538.

PR #538, opened 2024-10-30
util: change some layers to require recorders that are `Sync`

tobz:
## Context

Mentioned in #533, our recent change to require the global recorder to be `Sync` had a downstream effect for users taking advantage of certain layers like `Fanout` and `Router`. This is due to the fact that those layers only require taking something that is `Recorder`, which then is held as a `Box<dyn Recorder>`.

## Solution

This PR updates `Fanout` and `Router` (and by extension, their builders, `FanoutBuilder` and `RouterBuilder`) to take recorders that are `Sync`, and then holds them as `Box<dyn Recorder + Sync>`. This facilitates being able to once again use these layers as part of the global recorder.

The original intent, before the change to require `Sync` in the global recorder, was always to have layers be usable for the global recorder, so this change is simply aligning these utilities to match.

While there _are_ areas where a `Sync` recorder is not required, such as `metrics::with_local_recorder` or `metrics::set_default_local_recorder`... they're much more rare and generally scoped to testing, where something like `DebuggingRecorder` is being used and there would be no reason to use a layer. If somehow it turns out that there's a need for non-`Sync` layers for these use cases, we can address this by creating `Unsync` variants in the future.

### DIFF
--- a/metrics-util/CHANGELOG.md
+++ b/metrics-util/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- `FanoutBuilder` and `RouterBuilder` now both require recorders to be `Sync` to facilitate usage with being installed
+  as the global recorder.
+
 ## [0.18.0] - 2024-10-12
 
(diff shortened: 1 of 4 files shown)


tobz:
Released as `metrics-util@v0.19.0`.

Issue #540, opened 2024-11-02
Does metrics-exporter-prometheus support remote write feature?


taikulawo:
Hi, We are multiple process arch app. If many process bind to same prometheus server port, OS will report port already in use.
https://prometheus.io/docs/specs/remote_write_spec/

Our infra team don't want to deploy push_gateway, so we have to send metrics manually.

tobz:
👋🏻 

`metrics-exporter-prometheus` currently does not support the Remote Write specification. I'd be open to reviewing a PR adding support for this feature.

taikulawo:
wow，thankyou

PR #542, opened 2024-11-03
Feat: metric-explorers-prometheus add remote write

taikulawo:
https://prometheus.io/docs/specs/remote_write_spec/

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/remote_write_proto.rs
+++ b/metrics-exporter-prometheus/src/exporter/remote_write_proto.rs
@@ -1,6 +1,7 @@
 //! Types and utilities for calling Prometheus remote write API endpoints.
-
-use hyper::{header, Method, Request, Uri};
+// Copy from https://github.com/theduke/prom-write
+use http_body_util::Full;
+use hyper::{body::Bytes, header, Method, Request, Uri};
 
 /// Special label for the name of a metric.
 pub const LABEL_NAME: &str = "__name__";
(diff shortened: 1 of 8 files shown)


taikulawo:
close https://github.com/metrics-rs/metrics/issues/540

metrics-exporter-prometheus/src/exporter/remote_write_proto.rs
```diff
@@ -0,0 +1,274 @@
...
+                        .collect::<Vec<_>>();
+
+                    TimeSeries { labels, samples: vec![] }
+                });
+
+                let value = match sample.value {
+                    prometheus_parse::Value::Counter(v) => v,
+                    prometheus_parse::Value::Gauge(v) => v,
+                    prometheus_parse::Value::Histogram(_) => {
+                        Err("histogram not supported yet".to_string())?
```
taikulawo, line 99: @tobz I'm not sure how to process histogram and summary

tobz (changes requested):
Well that was fast. :)

The comment about not knowing how to handle histograms/summaries highlights what I see as the biggest thing that needs to change with this PR: we shouldn't be generating the exposition format only then to parse it back into an intermediate representation only to then encode it a second time in the Remote Write format.

We should simply be writing a new method on `Inner` that generates the encoded/compressed payload similar to how `Inner::render` does it, by evaluating the individual counters, gauges, and histograms.

taikulawo:
I don't know too much inner detail about metrics. I no idea how to build from inner type. But other codes should correct.

PR #549, opened 2025-01-12
docs: remove redundant paragraph

catbrained:
It looks like there is a redundant paragraph in the documentation explaining the same thing twice.

I removed the paragraph, and also a redundant word in the paragraph above.


tobz (approved):
Nice catch!

Thanks for cleaning this up. ❤️

PR #555, opened 2025-02-04
feature: add global labels support for dogstatsd exporter

### LINKED ISSUE #554: (feat-req) DogstatsdExporter: Support global labels
We are using http://github.com/dialtone/metrics-exporter-dogstatsd as our metrics exporter. 
We were little worried cause it was not maintained for sometime.

I am happy to see metrics team has taken http://github.com/dialtone/metrics-exporter-dogstatsd into official repo.

The old version 0.8.0 supports adding global tags to the exporter, which will be added to all the metrics. This is quite useful to add tags with server ids, from different servers.

When I tried to update to the latest one, I found the current version 0.9.0 is missing this feature.

Would you consider adding the support for global tags ?

And Thank you very much for adapting the dogstatsd exporter to official repo.

nappairam:
Fixes #554 

I took a stab at implementing this feature

The older version supports adding tags one by one. I took the liberty to make it a Vector with one api `with_global_labels` since it integrates well with the rest of the exporter.

Let me know your thoughts. I will update tests if this approach seems fine


nappairam:
@tobz Gentle reminder to review

metrics-exporter-dogstatsd/src/state.rs
```diff
@@ -96,7 +102,13 @@ impl State {
 
             active_counters += 1;
 
-            let result = writer.write_counter(&key, value, self.get_aggregation_timestamp());
+            let prefix = if key.name().starts_with("datadog.dogstatsd.client") {
+                &None
+            } else {
+                &self.config.global_prefix
+            }.as_deref();
```
tobz, line 109: I think you can make this cleaner:

```suggestion
            let prefix = if key.name().starts_with("datadog.dogstatsd.client") {
                None
            } else {
                self.config.global_prefix.as_deref()
            };
```

metrics-exporter-dogstatsd/src/state.rs
```diff
@@ -27,6 +27,12 @@ pub struct StateConfiguration {
 
     /// Whether or not to emit histograms as distributions.
     pub histograms_as_distributions: bool,
+
+    /// Global labels to add to all metrics
+    pub global_labels: Arc<Vec<Label>>,
```
tobz, line 32: `StateConfiguration` isn't `Clone`, so I don't think there's any benefit to wrapping the labels in an `Arc<T>`. We can just immutably borrow them when using them with `write_counter`, etc.

metrics-exporter-dogstatsd/src/writer.rs
```diff
@@ -121,21 +123,25 @@ impl PayloadWriter {
         true
     }
 
-    fn write_trailing(&mut self, key: &Key, timestamp: Option<u64>) {
-        write_metric_trailer(key, timestamp, &mut self.buf, None);
+    fn write_trailing(&mut self, key: &Key, timestamp: Option<u64>, labels: Arc<Vec<Label>>) {
```
tobz, line 126: Change `Arc<Vec<Label>>` to `&[Label]` just to be a little more generic, and probably call this `global_labels` just to be a little clearer that it's not the primary set of labels being written out.

Ditto this for every instance/method in the remainder of the file where take in the global labels.

tobz (changes requested):
Apologies on the delay for getting around to reviewing. :)

Overall, this looks reasonable. I let some feedback to clean things up, and I think we also need two new unit tests in the writer code for ensuring we properly write out the prefix when present, and that we handle global labels.

nappairam:
Thanks @tobz for review. I have updated the PR with the suggested changes and updated the unit tests
Please take a look again

nappairam:
@tobz Gentle reminder

tobz (approved):

PR #563, opened 2025-03-05
dogstatsd: reset client state on connection failure

tobz:
## Context

When the synchronous forwarder tries to transition from a disconnected to a connected state, it creates a new `Client` via `Client::from_forwarder_config`, which creates the underlying socket, connects to the remote server, and configures things like write timeouts.

In the case of UDP, "connect" is sort of a misnomer because UDP is connectionless... and so in case of trying to "connect" to a UDP address where nothing is listening, we only get an error to that effect when we try to _send_ on the socket. For UDS, however, the connection error comes immediately within `Client::from_forwarder_config`.

Unfortunately, the current code did not properly handle this case, which left the forwarder client in an inconsistent state: literally, `ClientState::Inconsistent`.

## Solution

This PR updates `ClientState::try_send` to properly reset the state back to "disconnected" when we fail to create our `Client`.

Prior to this PR, trying to connect to a non-existent UDS socket -- datagram or stream -- would immediately panic when trying to send the first payload. With this PR, it properly manifests as a connection error.

### DIFF
--- a/metrics-exporter-dogstatsd/src/forwarder/sync.rs
+++ b/metrics-exporter-dogstatsd/src/forwarder/sync.rs
@@ -91,7 +91,7 @@ impl ClientState {
                         *self = ClientState::Disconnected(config);
                         return Err(e);
                     }
-                }
+                },
                 ClientState::Ready(config, mut client) => {
                     let result = client.send(payload);
                     if result.is_ok() {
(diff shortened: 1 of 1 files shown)


tobz:
Released as `metrics-exporter-dogstatsd@v0.9.1`.

PR #565, opened 2025-03-17
make key comparison be order-independent

### LINKED ISSUE #558: Dimensions are order-sensitive
This creates 2 different gauges, which is probably not what you want, since the order of the dimensions differs between the two:

```rust
            gauge!("my_gauge", "foo" => "0", "bar" => "1").increment(1);
            gauge!("my_gauge", "bar" => "1", "foo" => "0").increment(1);
```

I saw this happen when using a HashMap for dimensions, and the HashMap getting a different order every time. This behavior should at least be better documented.

arielb1:
Fixes #558 

This has some negative performance impact. I can try caching the order in the key for keys with a sane number of dimensions, like the hash is currently cached, to try to reduce the impact. Do you think that belongs in this PR?


tobz:
I'm going to try and take a look at this sometime this week.

I'm not a super fan of _how_ much code is there, so I gotta see what the performance changes look like as-is vs any other potential approach.

metrics/src/key.rs
```diff
@@ -175,7 +175,34 @@ fn generate_key_hash(name: &KeyName, labels: &Cow<'static, [Label]>) -> u64 {
...
+            }
+        }
+        n => {
+            let mut labels_sort_map: Vec<usize> = (0..n).collect();
+            labels_sort_map.sort_by_key(|i| labels[*i].key());
+            for i in 0..n {
+                labels[labels_sort_map[i]].hash(state)
+            }
+        }
+    }
```
tobz, line 205: I wonder if there's a way we could wrap this loop unrolling into a helper function or something, since we effectively repeat it three times: here, the `PartialEq` impl, and the `Ord` impl.

I guess the biggest thing is that this impl only deals with one set of labels, while the other two trait impls deal with a pair of labels.

tobz (commented):
Overall, I can appreciate this change in spirit, although I've still yet to dig in locally to try to better grok what this might mean for performance. It looks like we don't already have a microbenchmark that really exercises this, unfortunately.

metrics/src/key.rs
arielb1: > I wonder if there's a way we could wrap this loop unrolling into a helper function or something, since we effectively repeat it three times: here, the PartialEq impl, and the Ord impl.

I can't come up with a good way of doing it.

tobz:
@arielb1 Nope, time is just at a premium these days so sometimes things take a little bit longer. 😅

tobz:
Released as part of `metrics@0.24.2`.

PR #568, opened 2025-03-25
fix(changelog): correct url

billf:


tobz:
😅 

Thanks for catching this!

tobz (approved):

PR #569, opened 2025-03-27
Fix unix datagram forwarder in metrics-exporter-dogstatsd

fuchsnj:
Two issues are fixed here:
1. `#[cfg(target_os = "linux")]` was used for unix sockets, which prevents other unix OSes like MacOS from using unix sockets.
2. The Unix datagram forwarder was binding to the address, which will fail since the Datadog Agent is expected to already be bound to that address as the server. This was switched to `unbound` since it is send-only.

### DIFF
--- a/metrics-exporter-dogstatsd/src/builder.rs
+++ b/metrics-exporter-dogstatsd/src/builder.rs
@@ -468,7 +468,7 @@ mod tests {
         );
     }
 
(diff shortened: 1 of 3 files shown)


tobz (approved):
LGTM 👍🏻

tobz:
Released as `metrics-exporter-dogstatsd@v0.9.2`.

PR #571, opened 2025-03-27
Remove unix datagram length-prefix for `metrics-exporter-dogstatsd`

fuchsnj:
The protocol for unix datagram should be the same as UDP, no length prefix. The prefix is only needed for "stream" sockets.


tobz (approved):

tobz:
Released as `metrics-exporter-dogstatsd@v0.9.3`.

PR #574, opened 2025-04-07
Support customizing Prometheus pushgateway HTTP method

trxcllnt:
This PR allows customizing the HTTP method used for the Prometheus pushgateway HTTP requests.

With this PR, a user can use [Vector's](https://vector.dev) `prometheus_pushgateway` sink (which [only supports HTTP POST](https://vector.dev/docs/reference/configuration/sources/prometheus_pushgateway/#http-methods-post-vs-put) requests) to translate between metrics-rs's pushgateway format and the native Prometheus remote-write format:

```yaml
# vector.yaml
sources:
  pushgateway:
    type: prometheus_pushgateway
    address: 0.0.0.0:9091
    aggregate_metrics: false
sinks:
  prometheus:
    type: prometheus_remote_write
    inputs:
      - pushgateway
    endpoint: http://prometheus:9090/api/v1/write
```

This works around the need for direct remote-write support requested in https://github.com/metrics-rs/metrics/issues/540.


metrics-exporter-prometheus/src/exporter/builder.rs
```diff
@@ -125,6 +125,7 @@ impl PrometheusBuilder {
         interval: Duration,
         username: Option<String>,
         password: Option<String>,
+        http_method: Option<hyper::Method>,
```
tobz, line 128: I'd prefer seeing this be `use_post: bool`. I'm not a huge fan of adding another parameter that everyone has to care about even if they just want the default... but I'd prefer seeing this be `use_post: bool`.

Otherwise, this is unconstrained and means that the user can screw up and pass a method that isn't valid (`GET`, etc) and that they actually need to import the `hyper::Method` type in the first place.

tobz (changes requested):

tobz (approved):

tobz:
Released as part of `metrics-exporter-prometheus@0.17.0`.

PR #576, opened 2025-04-15
prometheus: run metrics render in a tokio blocking thread

Tuetuopay:
Applications with large traffic and large cardinality can expect to have a lot of metrics to render. All the rendering code is sync, never yielding to runtime even once. For such large applications, the impact is clear: the tokio runtime has one of its worker thread blocked for as long as the rendering takes. The net result is latency spikes every time the process is scraped (we observed up to 15s for large production apps).

This has an additional footgun: the exporter can be installed from within a Tokio runtime, using it, or from outside of one, creating its own in a dedicated thread. This is not clear when using the exporter.

The code itself does not lend itself well to cooperative multitasking, relying on pure sync constructs, preventing an actual yield implementation.

Thus this commit moves the rendering to a tokio blocking thread, and fixes the exporter for everyone. Async stuff stays in tokio worker threads, and CPU-intesive work gets sent to dedicated threads.

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/http_listener.rs
+++ b/metrics-exporter-prometheus/src/exporter/http_listener.rs
@@ -2,7 +2,7 @@ use std::net::SocketAddr;
 
 use http_body_util::Full;
 use hyper::{
-    body::{self, Bytes, Incoming},
+    body::{Bytes, Incoming},
     header::{HeaderValue, CONTENT_TYPE},
     server::conn::http1::Builder as HyperHttpBuilder,
     service::service_fn,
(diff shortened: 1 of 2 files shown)


tobz:
Yeah, this is a reasonable change. 👍🏻

tobz:
Probably. Could be a follow-up PR, though.

Tuetuopay:
@trxcllnt most definitely! I'll do the same change.

Tuetuopay:
Done, and rebased on main.

tobz (approved):

tobz:
Looks good to me. 👍🏻

tobz:
Released as part of `metrics-exporter-prometheus@0.17.0`.

PR #578, opened 2025-05-09
fix(metrics-exporter-prometheus): :bug: allow usage in wasm

sfroment:
if you compile with cargo build --target wasm32-unknown-unknown compile will fail.
This PR Just aim to add compatibility for that to be able to use push gateway from a browser


tobz:
👋

I haven't reviewed this in depth, but at first glance, this looks like way more complexity than I'm willing to add just to support WebAssembly.

There's too much duplication and cfg flagging. I'd suggest trying to focus on cleaning that stuff up first, perhaps by trying to abstract the differing logic behind helper functions / types where necessary.

PR #582, opened 2025-05-17
fix: Fixed Prometheus export unit formatting

### LINKED ISSUE #553: Fix the appending of unit suffix for histogram metrics in Prometheus exporter.
In #535, support was added to `metrics-exporter-prometheus` for suffixing metrics based on their configured unit. This was applied uniformly to counters, gauges, and histograms, and relied on a change to one of the core functions used to write out a metric line in the exposition format.

#552 was raised to highlight a bug with how that change was gated, which made me realize that we're likely suffixing histograms wrong when the feature is enabled. For counters and gauges, the suffixing we do is fine: `my_counter` turns into something like `my_counter_seconds`, and so on. For histograms, however, lines like `my_histogram_sum` turn into `my_histogram_sum_seconds`.

As I understand it, at least, this is wrong, since the intent is that Prometheus would expect the base metric name to be turned into `<base>_bucket`, `<base>_sum`, and `<base>_count`... whereas we would be doing something like `<base>_bucket_seconds`, `<base>_sum_seconds`, and `<base>_count_seconds`. While the suffixing is uniform, Prometheus wants (again, as I understand it) a histogram metric to have the `_bucket`/`_sum`/`_count` suffixes so that it knows how to transforms queries against `<base>` into the right parts when doing histogram-y things.

Looking at the code, this _should_ be as simple as reordering how the unit suffix and the generic suffix (the parameter we use to add `_bucket`/`_sum`/`_count` in the first place) are applied.

ranger-ross:
This commit fixes improperly formatted metrics when `PrometheusBuilder::new().set_enable_unit_suffix(true)` is enabled.

Previously histogram metrics would have been formatted as:
* `<name>-sum-<unit>`
* `<name>-count-<unit>`
* `<name>-bucket-<unit>`

However the correct formatting is:
* `<name>-<unit>-sum`
* `<name>-<unit>-count`
* `<name>-<unit>-bucket`

You can see some example metric names in the Prometheus [histogram best practices page](https://prometheus.io/docs/practices/histograms/)

fixes #553

### DIFF
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -77,14 +77,8 @@ pub fn write_metric_line<T, T2>(
 
     match unit {
         Some(Unit::Count) | None => {}
-        Some(Unit::Percent) => {
-            buffer.push('_');
-            buffer.push_str("ratio");
-        }
-        Some(unit) => {
-            buffer.push('_');
-            buffer.push_str(unit.as_str());
-        }
+        Some(Unit::Percent) => add_unit(buffer, "ratio"),
+        Some(unit) => add_unit(buffer, unit.as_str()),
     }
 
     if !labels.is_empty() || additional_label.is_some() {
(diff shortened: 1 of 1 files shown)


tobz (changes requested):
Sorry, I thought I had submitted this review but just realized I left it pending... 😓

tobz (approved):

ranger-ross:
@tobz When convenient, would you mind cutting a new release with this patch  🙏

Issue #584, opened 2025-05-27
Remaining std AtomicU64 in metrics-util


opkaizen:
Hello, 

After moving away from `std::atomic::AtomicU64` for 32bit compatility (https://github.com/metrics-rs/metrics/issues/323), one remains in `metrics-util/src/storage/reservoir.rs`, making this crate impossible to build in 32 bit environments.

The following patch seems to work, but I'm not sure yet of the implications it would have:
```
diff --git a/metrics-util/src/storage/reservoir.rs b/metrics-util/src/storage/reservoir.rs
index 1b8e820..5a7f38a 100644
--- a/metrics-util/src/storage/reservoir.rs
+++ b/metrics-util/src/storage/reservoir.rs
@@ -1,10 +1,11 @@
 //! An atomic sampling reservoir.
 
+use metrics::atomics::AtomicU64;
 use std::{
     cell::UnsafeCell,
     sync::{
         atomic::{
-            AtomicBool, AtomicU64, AtomicUsize,
+            AtomicBool, AtomicUsize,
             Ordering::{Acquire, Relaxed, Release},
         },
         Mutex,
```

tobz:
Hmmm, I think that should be fine? At least if it compiles and things seem to continue working.


Want to throw up a PR with that change?

PR #585, opened 2025-05-27
fix(metrics-util): Replace std AtomicU64 with metrics

opkaizen:
Fixes https://github.com/metrics-rs/metrics/issues/584


tobz:
Correction: this is now released as `metrics-util@v0.19.2`. I made a mistake when cutting the release by including a breaking change, so I'll be yanking that shortly, and `metrics-util@v0.20.0` will replace it.

PR #586, opened 2025-06-01
fix(dogstatsd): properly pack multiple metrics into each payload up to the configured limit

tobz:
## Context

Prior to this PR, the writer logic ensured we didn't write metrics that exceeded the maximum payload size... but it didn't actually pack multiple metrics into a payload. This is inefficient because we often have the opportunity to send many metrics in a single write operation while still obeying the maximum payload limit.

This PR refactors `PayloadWriter` to properly pack multiple metrics into each payload up to the configured limit. We did a lot of deduplication and overly-verbose method naming as the code can otherwise be a little terse/hard to easily understand. We might want to revisit the multiple buffers approach in the future, but for right now, this should be good enough while still getting the job done.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1518,9 +1518,9 @@ dependencies = [
 
 [[package]]
 name = "predicates"
-version = "3.1.3"
+version = "3.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


tobz:
Released as `metrics-exporter-dogstatsd@v0.9.5`.

PR #588, opened 2025-06-17
"Consume" counters and gauge on snapshot

villainb-dg:
## Summary

This commit modifies the behavior of the snapshot to consume (or reset) the observed counters and gauges. The motivation is to make them behave in a similar way to histograms whose values are cleared from the backing storage during snapshot. The counters and gauges are set to 0 as it is their initial values at creation time.

## Tests

[x] unit tests

### DIFF
--- a/metrics-util/src/debugging.rs
+++ b/metrics-util/src/debugging.rs
@@ -105,11 +105,14 @@ impl Snapshotter {
 
         for (ck, _) in seen.into_iter() {
             let value = match ck.kind() {
-                MetricKind::Counter => {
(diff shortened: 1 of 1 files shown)


anson-vandoren:
This is quite the breaking change for a point release...

tobz:
Hmmm, true. This should probably be a new minor release.

I'll redo the release soon.

tobz:
This is now released as `metrics-util@v0.20.0`, and I've yanked `v0.19.2`.

Thanks @anson-vandoren for pointing out that issue. 👍🏻

PR #590, opened 2025-07-03
metrics-exporter-prometheus: Add API to initialize HTTP listener from existing TcpListener

michelhe:
This commit adds a new API to the `PrometheusBuilder` that allows users to provide their own configured TCP listener. The use case is to enable users to configure the TCP listener before passing it to the exporter - for example, to set socket options or bind to port 0 and retrieve the port number assigned by the OS.

My own use case for this is to use port=0 and then grab the port, log it and then give it to `PrometheusBuilder`

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/builder.rs
+++ b/metrics-exporter-prometheus/src/exporter/builder.rs
@@ -105,7 +105,12 @@ impl PrometheusBuilder {
         self
     }
 
-    /// Same as `with_http_listener`, but allows to bring your own configured tcp listener.
+    /// Configures the exporter to expose an HTTP listener using an existing TCP listener.
+    ///
+    /// This allows for customizing the underlying TCP socket, or binding to an ephemeral
+    /// port and capturing the port before passing off the listener.
+    ///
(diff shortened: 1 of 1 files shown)


tobz:
I do intend to look at this soon, but changes to `PrometheusBuilder` that add additional variants are usually low down on my list: it already contains a large number of methods, which makes trying to thinking through if we really should add something take a little more time.

metrics-exporter-prometheus/src/exporter/builder.rs
```diff
@@ -105,6 +105,25 @@ impl PrometheusBuilder {
         self
     }
 
+    /// Same as `with_http_listener`, but allows to bring your own configured tcp listener.
```
tobz, line 108:
```suggestion
    /// Configures the exporter to expose an HTTP listener using an existing TCP listener.
    ///
    /// This allows for customizing the underlying TCP socket, or binding to an ephemeral
    /// port and capturing the port before passing off the listener.
    ///
    /// See [`with_http_listener`] for more details.
```

PR #593, opened 2025-07-07
fix(dogstatsd): Fix panic when using length-prefixed streams

### LINKED ISSUE #592: dogstatsd - Metrics forwarder thread panics in UDS stream mode
This issue was introduced recently in #586 - I observed that whenever I'm using dogstatsd recorder with a unix tcp socket (`unix://<path>`) the forwarder thread would panic in debug mode with `attempt to subtract with overflow` in debug mode, and in release mode would panic elsewhere when attempting to `u32::try_from` from an overflow value.

After some debugging, I found the root cause is that the new `ConsumingBufferSwap` logic removes the pre-allocated 4 bytes of length prefix which is only used in unix mode, breaking the buffer. 

I will be sending a PR to fix shortly.

The simplest way for reproducing this is simply writing an example binary and try running it (the socket does not have to exist)

```
use std::time::Duration;

use metrics_exporter_dogstatsd::DogStatsDBuilder;
use tracing_subscriber;


fn main() {
    // Install a global subscriber to capture logs..
    tracing_subscriber::fmt().try_init().unwrap();

    let recorder = DogStatsDBuilder::default()
        .with_flush_interval(Duration::from_secs(1))
        .with_remote_address("unix:///tmp/dsd_test.sock").unwrap().build().unwrap();
    metrics::set_global_recorder(recorder).unwrap();

    // Wait for metrics forwarder to flush..
    for _ in 0..10000 {
        metrics::counter!("test_1").increment(1);
        std::thread::sleep(std::time::Duration::from_millis(10));
    }

    // The thread should have panicked in the background..
}
```

To reproduce this bug in unit test on master, simply apply the following patch and run the `length_prefix` test:
```git
diff --git a/metrics-exporter-dogstatsd/src/writer.rs b/metrics-exporter-dogstatsd/src/writer.rs
index 591195a..81a0604 100644
--- a/metrics-exporter-dogstatsd/src/writer.rs
+++ b/metrics-exporter-dogstatsd/src/writer.rs
@@ -958,6 +958,11 @@ mod tests {
 
             let actual = buf_from_writer(&mut writer);
             assert_eq!(actual, expected);
+
+            // --- The below code panics due to ConsumingBufferSwap wiping the length prefix bug. ---
+            // Write another payload
+            let result = writer.write_distribution(&key, values.iter().copied(), None, None);
+            assert_eq!(result.payloads_written(), 1);
         }
     }
 

 ```

And we'll be greeted with the following backtrace:
```
RUST_BACKTRACE=1 cargo test -p metrics-exporter-dogstatsd length_prefix
    Finished test [unoptimized + debuginfo] target(s) in 0.04s
     Running unittests src/lib.rs (target/debug/deps/metrics_exporter_dogstatsd-926db075ee543c8d)

running 1 test
test writer::tests::length_prefix ... FAILED

failures:

---- writer::tests::length_prefix stdout ----
thread 'writer::tests::length_prefix' panicked at metrics-exporter-dogstatsd/src/writer.rs:166:9:
attempt to subtract with overflow
stack backtrace:
   0: rust_begin_unwind
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:127:5
   3: metrics_exporter_dogstatsd::writer::PayloadWriter::current_payload_len
             at ./src/writer.rs:166:9
   4: metrics_exporter_dogstatsd::writer::PayloadWriter::commit
             at ./src/writer.rs:234:35
   5: metrics_exporter_dogstatsd::writer::PayloadWriter::try_write_multiple
             at ./src/writer.rs:419:16
   6: metrics_exporter_dogstatsd::writer::PayloadWriter::write_distribution
             at ./src/writer.rs:498:9
   7: metrics_exporter_dogstatsd::writer::tests::length_prefix
             at ./src/writer.rs:964:26
   8: metrics_exporter_dogstatsd::writer::tests::length_prefix::{{closure}}
             at ./src/writer.rs:926:24
   9: core::ops::function::FnOnce::call_once
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/ops/function.rs:250:5
  10: core::ops::function::FnOnce::call_once
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    writer::tests::length_prefix

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 23 filtered out; finished in 0.02s

error: test failed, to rerun pass `-p metrics-exporter-dogstatsd --lib`
```

michelhe:
Resolves #592 

The internal payload writer incorrectly consumed the 4-byte length prefix
required for Unix stream framing. When the payload buffer was drained,
these prefix bytes were removed, causing an arithmetic overflow in a
subsequent payload length calculation.

This fix simplifies the implementation by removing the complex
`ConsumingBufferSwap` helper. The `Payloads` iterator now drains
consumed metrics directly from the buffer in-place. By draining only the
necessary bytes, it preserves the trailing length prefix and ensures
the writer's state remains consistent.

### DIFF
--- a/metrics-exporter-dogstatsd/src/writer.rs
+++ b/metrics-exporter-dogstatsd/src/writer.rs
@@ -958,6 +958,11 @@ mod tests {
 
             let actual = buf_from_writer(&mut writer);
             assert_eq!(actual, expected);
+
+            // --- The below code panics due to ConsumingBufferSwap wiping the length prefix bug. ---
+            // Write another payload
+            let result = writer.write_distribution(&key, values.iter().copied(), None, None);
+            assert_eq!(result.payloads_written(), 1);
         }
     }
 
--- a/metrics-exporter-dogstatsd/src/writer.rs
+++ b/metrics-exporter-dogstatsd/src/writer.rs
@@ -1,7 +1,4 @@
-use std::{
-    ops::{Deref, DerefMut},
-    vec::Drain,
-};
+use std::vec::Drain;
 
 use metrics::{Key, Label};
 
@@ -523,16 +520,26 @@ impl PayloadWriter {
 }
 
 /// Iterator over all payloads written by a `PayloadWriter`.
+///
+/// The source payload buffer is immediately drained of consumed data during the creation of this iterator
+/// ("pre-pooping our pants"). This ensures that the end state - the payload buffer contains only preserved
+/// bytes (like length prefixes) - is established immediately.
+///
+/// https://faultlore.com/blah/everyone-poops/
 pub struct Payloads<'a> {
-    payloads_buf: ConsumingBufferSwap<'a, u8>,
+    payloads_buf: Vec<u8>,
     start: usize,
     offsets: Drain<'a, usize>,
 }
 
 impl<'a> Payloads<'a> {
     fn new(payload_buf: &'a mut Vec<u8>, offsets: &'a mut Vec<usize>) -> Self {
+        // When draining payloads, we need to preserve any bytes that come after the last offset.
+        // These bytes (like length prefixes) are not part of the current payloads but are needed
+        // for the next write operation.
+        let drain_size = offsets.last().copied().unwrap_or(0);
         Self {
-            payloads_buf: ConsumingBufferSwap::new(payload_buf),
+            payloads_buf: payload_buf.drain(0..drain_size).collect(),
             start: 0,
             offsets: offsets.drain(..),
         }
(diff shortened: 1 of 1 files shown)


tobz (approved):
Nice catch. 👍🏻 

I don't _love_ the additional allocation every time we consume the payloads, and I think we can do better by using `Drain<'a, T>` directly along with some refactoring of when we write the length prefix placeholder... but panics are bad so let's get this merged as-is to solve the immediate problem.

michelhe:
> Going to merge as-is even with the failing Clippy check since that can be fixed in a follow-up.

Sorry for forgetting to re-run clippy :( 
I actually tried to solve the allocations nicely with `Drain<'a, T>` but it ended up too big of a refactor, so I went with this one so I can use the crate asap.

tobz:
Released as `metrics-exporter-dogstatsd@v0.9.6`.

Thanks for your contribution! ❤️

PR #596, opened 2025-07-17
Prometheus: Add setting for recommended metric names

michelhe:
This commit adds a setting to PrometheusBuilder to add the '_total' suffix to counter names, compatible with [Prometheus naming best practices](https://prometheus.io/docs/practices/naming/).

This is useful for migrating from the [prometheus-client] crate, (where the counter suffix is enforced) without breaking existing dashboards.

[prometheus-client](https://docs.rs/prometheus-client/latest/prometheus_client/)

### DIFF
--- a/metrics-exporter-prometheus/src/exporter/builder.rs
+++ b/metrics-exporter-prometheus/src/exporter/builder.rs
@@ -49,6 +49,7 @@ pub struct PrometheusBuilder {
     recency_mask: MetricKindMask,
     global_labels: Option<IndexMap<String, String>>,
     enable_unit_suffix: bool,
+    enable_compat_counter_names: bool,
 }
 
 impl PrometheusBuilder {
@@ -82,6 +83,7 @@ impl PrometheusBuilder {
             recency_mask: MetricKindMask::NONE,
             global_labels: None,
             enable_unit_suffix: false,
+            enable_compat_counter_names: false,
         }
     }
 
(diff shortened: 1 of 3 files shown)


tobz:
👋🏻 

I'd rather see this updated here: https://github.com/metrics-rs/metrics/blob/5bf2aee5318715bef95a0d6dacc63fa3a31541bb/metrics-exporter-prometheus/src/formatting.rs#L79

Having _two_ distinct settings for controlling metric name suffixes is too much.

michelhe:
> 👋🏻
> 
> I'd rather see this updated here:
> 
> https://github.com/metrics-rs/metrics/blob/5bf2aee5318715bef95a0d6dacc63fa3a31541bb/metrics-exporter-prometheus/src/formatting.rs#L79
> 
> Having _two_ distinct settings for controlling metric name suffixes is too much. I think we would just add this for the `Unit::Count` case, though.

Only `Unit::Count` isn't enough to allow migration from [prometheus-client](https://docs.rs/prometheus-client/latest/prometheus_client/) . That crate formats unit-less counters with `_total` suffix and counters with unit information with the `_{unit}_total` suffix, following the [Prometheus naming best practices](https://prometheus.io/docs/practices/naming/) document to the word.

See these lines from the above link:
> http_requests_total (for a unit-less accumulating count)
 process_cpu_seconds_total (for an accumulating count with unit)

If you don't want yet another setting, maybe we can have a `set_prometheus_client_compat_mode`  that implictly enables the unit suffix as well?

michelhe:
I pushed the changes you suggested, and also had to fix some stuff in `formatting.rs`.

Mainly, the fix I made in 80960de could be breaking existing metric names for users who use unit-suffixes :/

tobz:
I need to refresh my memory on this but the existing support for adding a suffix in `write_metric_line` was just to add stuff for aggregated histograms/summaries (e.g., `my_metric_name_sum`, etc) so ... I think it may actually be _broken_ as-is if people are using unit suffixes. 🤔 

(The idea being that your change is now fixing it.)

PR #597, opened 2025-07-23
fix: ensure counter metrics include _total suffix in HELP and TYPE lines

michelhe:
Following #596 

The Prometheus exporter now properly appends the _total suffix to counter metric names in both help text and type declarations, ensuring consistency with Prometheus naming conventions.

This also makes the python `prometheus_client.parser.text_string_to_metric_families` parse the metrics better.

Even though not mandatory by the exposition format, I've seen the prometheus go client append these suffixes to both HELP and TYPE lines. For example see:

```bash
docker run -d --rm -p 9090:9090 prom/prometheus
curl http://localhost:9090/metrics
```

### DIFF
--- a/metrics-exporter-prometheus/src/formatting.rs
+++ b/metrics-exporter-prometheus/src/formatting.rs
@@ -29,13 +29,15 @@ pub fn key_to_parts(
 /// Writes a help (description) line in the Prometheus [exposition format].
 ///
 /// [exposition format]: https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md#text-format-details
-pub fn write_help_line(buffer: &mut String, name: &str, unit: Option<Unit>, desc: &str) {
+pub fn write_help_line(
+    buffer: &mut String,
+    name: &str,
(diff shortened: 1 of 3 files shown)


metrics-exporter-prometheus/src/formatting.rs
```diff
@@ -121,24 +114,58 @@ pub fn write_metric_line<T, T2>(
...
+/// Otherwise, we append the unit to the end of the metric name.
+fn add_unit_if_missing(buffer: &mut String, unit: Unit) {
+    const KNOWN_SUFFIXES: [&str; 4] = ["_sum", "_count", "_bucket", "_total"];
+
+    let unit = match unit {
+        Unit::Count => {
+            // For count, we don't suffix the unit.
+            return;
+        }
+        Unit::Percent => "ratio",
```
michelhe, line 151: @tobz I'm not sure where the "percent" -> "ratio" conversion was there originally.

tobz:
Released in `metrics-exporter-prometheus@v0.18.0`.

PR #598, opened 2025-07-31
histogram: change bucket behavior // comply with openmetrics

baloo:
OpenMetrics [defines][openmetrics-histogram] histograms as cumulative. A value should be assigned to a bucket and only one bucket, the sum of the bucket is expected to be the number of values recorded:

```
Histogram MetricPoints MUST have one bucket with an +Inf threshold. Buckets MUST be cumulative. As an example for a metric representing request latency in seconds its values for buckets with thresholds 1, 2, 3, and +Inf MUST follow value_1 <= value_2 <= value_3 <= value_+Inf. If ten requests took 1 second each, the values of the 1, 2, 3, and +Inf buckets MUST equal 10.
```

This obviously changes the behavior of the histogram and is a breaking change.

[openmetrics-histogram]: https://github.com/prometheus/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram


baloo:
This is obviously a breaking change, but I have a hard time tracking in the git history where that choice was made, I'm opening the PR more as way to open the discussion about that behavior.

At least in my tests, this makes the prometheus report "consistent" with what a grafana would expect to be able to display a proper latency histogram.

tobz:
I believe you've misread the specification. Admittedly, it's not written as well as it could be.

What it's indicating is that each bucket in the example must equal 10, because the recorded values are all less than or equal to the bucket's upper bound. It's not saying that the value of every bucket, when summed, must equal 10.

You can see this in action in Prometheus' own docs: https://prometheus.io/docs/instrumenting/exposition_formats/#text-format-example. Note the bucket values only go up as the upper bound (`le`) increases, and the sum is equal to the `+Inf` bucket, which would be impossible otherwise.

baloo:
Yeah, I've read it backwards. Sorry for the noise.
Thank you!

Issue #601, opened 2025-08-19
Support for optimized histogram representations (otel "ExponentialHistogram" or prometheus native histogram)


alex:
Current histogram aggregation has a bunch of downsides, which mostly revolve around accuracy, efficiency, and annoyance from bucket management.

OpenTelemetry has introduced a new ExponentialHistogram, and prometheus has an (experimental) native histogram. My understanding is that while these are not the same, their representations are convertible.

It'd be outstanding if metrics were to make these types available. I think there's an API question on whether the public API for these is to continue to use `histogram!` and then on the backend you configure that all your histograms are native ones, or if it's a separate public API for them.

tobz:
The public API for recording a histogram is always, always, going to be `histogram!`.

Choosing how to capture/render those histograms is purely an exporter concern. It should definitely be possible to update `metrics-exporter-prometheus` to support emitting histograms as native histograms instead of aggregated histograms.

alex:
Ok! So then this bug is: support configuring `metrics-exporter-prometheus` to use native histograms.

PR #602, opened 2025-08-19
Added support for the protobuf format for prometheus

alex:
this is a pre-req for #601, because native histograms are only supported with protobufs

in addition to the unit tests, I also ran the server example with a local prometheus configured to only talk protobuf and confirmed that metrics flowed.

unfortunately, this format is not well documented so I made some reasonable inferences, but wasn't positive about many things, so a close eye is appreciated!

### DIFF
--- a/metrics-exporter-prometheus/src/protobuf.rs
+++ b/metrics-exporter-prometheus/src/protobuf.rs
@@ -1,0 +1,301 @@
+//! Protobuf serialization support for Prometheus metrics.
+
+use indexmap::IndexMap;
+use metrics::Unit;
+use prost::Message;
+use std::collections::HashMap;
+
+use crate::common::Snapshot;
+use crate::distribution::Distribution;
+use crate::formatting::sanitize_metric_name;
+
+// Include the generated protobuf code
+mod pb {
+    #![allow(missing_docs, clippy::trivially_copy_pass_by_ref, clippy::doc_markdown)]
+    include!(concat!(env!("OUT_DIR"), "/io.prometheus.client.rs"));
+}
+
+#[cfg(feature = "http-listener")]
+pub(crate) const PROTOBUF_CONTENT_TYPE: &str =
+    "application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited";
+
+/// Renders a snapshot of metrics into protobuf format using length-delimited encoding.
+///
+/// This function takes a snapshot of metrics and converts them into the Prometheus
+/// protobuf wire format, where each `MetricFamily` message is prefixed with a varint
+/// length header.
+#[allow(clippy::too_many_lines)]
+pub(crate) fn render_protobuf(
+    snapshot: &Snapshot,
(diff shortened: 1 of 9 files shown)


alex:
Looks like there's some clippy issues, will fix those this evening!

alex:
Ok, clippy issues should be resolved!

alex:
Is a maintainer able to set the option on the repo so that new contributor's CI runs don't need to be explicitly approved?

metrics-exporter-prometheus/src/protobuf.rs
```diff
@@ -0,0 +1,327 @@
...
+    snapshot: &Snapshot,
+    descriptions: &HashMap<String, (metrics::SharedString, Option<Unit>)>,
+    global_labels: &IndexMap<String, String>,
+    counter_suffix: Option<&'static str>,
+) -> Vec<u8> {
+    let mut output = Vec::new();
+
+    // Process counters
+    for (name, by_labels) in &snapshot.counters {
+        let sanitized_name = sanitize_metric_name(name);
```
tobz, line 38: Just a _general_ note: we might need to/want to evaluate, as part of a future PR, relaxing this sanitization since I _think_ full UTF-8 metric names/tags are allowed for the Protobuf payloads? (and the text format, if you use a recent enough version and escape properly, yadda yadda)

alex: Presumably none of that escaping is required for the protobuf format. I'm happy to drop if you'd like? These lines were not added with a ton of thought :-/

tobz:
> Is a maintainer able to set the option on the repo so that new contributor's CI runs don't need to be explicitly approved?

It's explicitly set this way due to the potential for exploits to be invoked through automatically running CI against untrusted changes in PR. I can appreciate that it makes it harder to iterate on CI-related fixes, but I don't want to want to have to think any harder than absolutely necessary about repo security, so this is how it stays. 😅

metrics-exporter-prometheus/src/protobuf.rs
```diff
@@ -0,0 +1,327 @@
...
+pub(crate) const PROTOBUF_CONTENT_TYPE: &str =
+    "application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited";
+
+/// Renders a snapshot of metrics into protobuf format using length-delimited encoding.
+///
+/// This function takes a snapshot of metrics and converts them into the Prometheus
+/// protobuf wire format, where each `MetricFamily` message is prefixed with a varint
+/// length header.
+#[allow(clippy::too_many_lines)]
+pub(crate) fn render_protobuf(
```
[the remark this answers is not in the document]
alex: Done!

alex:
FWIW, based on my experience maintaining other OSS projects, I think the best way to handle that is to ensure that all jobs that run on pull_request envs don't require any secrets or do anything destructive, so that the worst thing anyone can do is waste resources.

PR #605, opened 2025-09-12
refactor how collected metrics are stored to avoid parsing labels

alex:
currently the prootbuf implementation takes the stringified-labels and parses them back into their original form, which is silly. refactored to avoid that, and in the process also avoid some copies and allocations.

this changes public APIs around these, so should only be landed with an appropriate version bump


alex:
great, thanks! will try to get the final (hopefully!) histogram PR up soon

PR #606, opened 2025-09-19
Fix atomicity violation in AtomicSamplingReservoir::consume

quantpoet:
Replaced separate atomic load/store operations with a single fetch_xor to eliminate 
the race condition where multiple threads could read the same value and write 
conflicting results. This ensures the read-modify-write operation is truly atomic 
and resolves the lockbud-detected atomicity violation.


tobz:
👋🏻 

Sorry about the delay.

tobz (changes requested):

tobz:
Released in `metrics-util@v0.20.1`.

PR #608, opened 2025-10-01
Upgrade hashbrown to 0.16

djc:
I can see that there's a Dependabot config, but the last actual Dependabot PR seems to be #544, from a year ago.

See here:

https://github.com/metrics-rs/metrics/pulls?q=is%3Apr+label%3Adependencies+sort%3Aupdated-desc

In any case, would be nice to keep dependencies up to date. I can probably take care of some of the other ones if this is merged:

- [ ] itertools
- [ ] mockall
- [ ] ordered-float
- [ ] prost
- [ ] radix_trie
- [ ] ratatui
- [ ] critertion (currently at `=0.3.3` -- not sure why?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,6 +567,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
+name = "foldhash"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


djc:
> Happy to accept a PR to update the ones mentioned above. I believe the pinning of `criterion` was related to a transient dependency it carries (or carried, at the time, at least) that was causing us to have to bump MSRV... which I refuse to do just for benchmarks.

I recommend using `cargo check --lib` in MSRV jobs for this exact reason.

tobz:
Yeah, go for it. 👍🏻

tobz:
Released in `metrics-util@v0.20.1`.

PR #610, opened 2025-10-05
fixes #601 -- implement native histogram support for metrics-exporter-prometheus

### LINKED ISSUE #601: Support for optimized histogram representations (otel "ExponentialHistogram" or prometheus native histogram)

alex:
this draws heavily from the Go and Python implementations -- its definitely possible that I missed some essential details, so I'd appreciate a critical review.

in local testing (using the provided example) with a prometheus server, it seems to work!


tobz:
Overall, this _seems_ reasonable to me. I gave it a reasonable review, and it all makes sense to me... but I don't really use Prometheus at all so more subtle issues/pitfalls may have very well eluded me during my review. 😅

alex:
Thanks for the review! I'll see if I can find any additional reviewers with deeper familiarity with prom native histograms.

tobz:
In the interest of getting this out there to be played with, I'm going to merge it as-is. Worst case, someone realizes it's broken and we revert it or fix it. 🤷🏻

PR #611, opened 2025-10-06
Remove rust-toolchain.toml file

djc:
Current Rust Analyzer does not support any compilers older than 1.78. It seems suboptimal to use the MSRV toolchain by default as it is prone to be slower and less reliable than newer compilers. Instead, the MSRV is checked explicitly in CI.


tobz:
Ah, that's a bummer about `rust-analyzer`.

I had pinned it to MSRV to try and reduce the toil of writing someone that would break the MSRV but only alert you when you got to CI... shorten the loop, basically.

If `rust-analyzer` doesn't support that old of a version, though, then might as well just lean on CI.

tobz (approved):

PR #635, opened 2025-11-09
fix std Cow conversion being Sized #581

### LINKED ISSUE #581: `SharedString` cannot be converted into `std::borrow::Cow<'static, str>'
Although the conversion is implemented, the compiler won't let me use it because `str` is not `Sized`.

For example, the following code:
```rust
struct MetricMetadata {
    description: SharedString,
    unit: Option<Unit>,
}

impl MetricMetadata {
    fn description(&self) -> std::borrow::Cow<'static, str> {
        self.description.into()
    }
}
```
refuses to compile with the following error:
```
error[E0277]: the size for values of type `str` cannot be known at compilation time
  --> src/metadata.rs:38:26
   |
38 |         self.description.into()
   |                          ^^^^ doesn't have a size known at compile-time
   |
   = help: the trait `std::marker::Sized` is not implemented for `str`
   = note: required for `std::borrow::Cow<'_, str>` to implement `std::convert::From<metrics::cow::Cow<'_, str>>`
   = note: required for `metrics::cow::Cow<'_, str>` to implement `std::convert::Into<std::borrow::Cow<'_, str>>`
```

Javyre:
fixes #581 

(would be nice to also have a way to turn KeyName into a Cow)

### DIFF
--- a/metrics/src/cow.rs
+++ b/metrics/src/cow.rs
@@ -324,7 +324,7 @@ impl<'a> From<std::borrow::Cow<'a, str>> for Cow<'a, str> {
     }
 }
 
-impl<'a, T: Cowable> From<Cow<'a, T>> for std::borrow::Cow<'a, T> {
+impl<'a, T: Cowable + ?Sized> From<Cow<'a, T>> for std::borrow::Cow<'a, T> {
     #[inline]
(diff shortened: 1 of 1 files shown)


cherryman:
@tobz any chance we can merge this tiny pr 🥺

PR #641, opened 2025-11-14
feat(prometheus): add render_to_write function

nmooscisco:
In order to enable callers to emit metrics in chunks, or to a small buffer to keep memory consumption flat, this patch introduces the `render_to_write` function, which can be used to stream metrics in chunks, avoiding buffering the entire set of metrics into a single string.


tobz:
Released in `metrics-exporter-prometheus@v0.18.0`.

PR #651, opened 2025-12-09
Suggestion: use rapidhash for KeyHasher

hoxxep:
Disclosure: I am the author of rapidhash.

Hey! This PR is a suggestion to replace the use of ahash with [rapidhash](https://github.com/hoxxep/rapidhash?tab=readme-ov-file#benchmarks). Rapidhash has been used in production workloads for the last year and outperforms ahash on almost all inputs and platforms.

This makes a significant improvement on the following benchmarks:
- macros/global_initialized/with_dynamic_labels (-30%)
- macros/local_initialized/with_dynamic_labels (-30%)
- registry/uncached op (basic) (-19%)
- registry/uncached op (labels) (-29%)
- registry/const key data overhead (labels) (-4%)
- registry/owned key overhead (basic) (-45%)
- registry/owned key overhead (labels) (-37%)

The improvements were benchmarked on an M1 Max with and without `-C target-cpu=native`. The [rapidhash detailed benchmarks](https://github.com/hoxxep/rapidhash/tree/master/docs) suggest rapidhash outperforms the ahash primary (with AES instructions) and fallback (without AES instructions) algorithms on almost all inputs on almost all platforms (1kB+ strings on x86 with AES enabled being a notable exception). I'm happy to talk through the benchmarks in more detail if that's helpful.

Note that metrics does not randomly seed the KeyHasher, and so the hashes are not DoS resistant in neither the old ahash implementation nor this PR. Let me know if this is a concern and I can adjust this PR to use rapidhash's [`GlobalState`](https://docs.rs/rapidhash/latest/rapidhash/fast/type.GlobalState.html) to provide some DoS resistance.

The rapidhash MSRV is 1.71.0, which should be enough for the current metrics MSRV.

Let me know if you have further questions about rapidhash, or would like me to assist in further benchmarking? Cheers!


tobz (approved):

tobz:
Released as `metrics@v0.24.4` and `metrics-util@v0.20.2`.

PR #658, opened 2026-01-12
perf: use write! macro in write_metric_line

DaniPopes:
Avoids intermediate String allocations from to_string() calls by writing directly to the buffer with the write! macro.


tobz (approved):

tobz:

PR #660, opened 2026-01-14
metrics-exporter-tcp: Decode all packets in receive buffer

laanwj:
Decode the entire receive buffer (as far as possible) when handling metrics, instead of only one packet per read. This prevents packets from queing up forever.

This change needs to be made both in `metrics-observer` and the example `tcp-client`.

Closes #659.


tobz (approved):
Nice, this would certainly explain the queueing behavior. 😅

tobz:
Released as `metrics-exporter-tcp@v0.11.2` and `metrics-observer@v0.5.1`.

PR #661, opened 2026-01-16
feat(prometheus): add render_protobuf_to_write function

nmooscisco:
In the same vein as #641, this patch introduces a new `render_protobuf_to_write` function, which can be used to stream metrics in chunks in protobuf format, avoiding buffering the entire set of metrics into a single byte buffer.


tobz:

PR #672, opened 2026-02-04
feat(prometheus): add lock-free descriptions map

nmooscisco:
With the recent addition of the streaming export methods, a deadlock risk has emerged where if an export is in progress, but stalls due to a slow HTTP client, and a new metric is being created, the modification of the descriptions map is starved. This is relatively easy to create by accident in a Tokio setting, where one Tokio task creates a new metric and another task, which gets scheduled to the same thread, is exporting metrics using the `render_to_write` family of functions in conjunction with `tokio_util`'s `simplex` + `SyncIoBridge` types.

In order to prevent this kind of problem by design, this patch changes the underlying storage to an `evmap`, which is an eventually-consistent single-writer, many-reader map. Upon export, a new read handle is cloned and is used as a snapshot. Writes are staged in serial, and then are committed upon completion of an export and during the upkeep phase. This mechanism prevents deadlocks by design, as a read-half lock is never held longer than it takes to perform a clone, and the write-half lock only needs to be held to stage and commit a write, which is only performed when it is safe to do so.

### DIFF
--- a/metrics-exporter-prometheus/Cargo.toml
+++ b/metrics-exporter-prometheus/Cargo.toml
@@ -43,6 +43,7 @@ _hyper-client = [
 base64 = { workspace = true }
 quanta = { workspace = true }
 thiserror = { workspace = true }
+evmap = { workspace = true }
 
 # Optional
 http-body-util = { workspace = true, optional = true }
--- a/metrics-exporter-prometheus/src/recorder.rs
+++ b/metrics-exporter-prometheus/src/recorder.rs
@@ -4,7 +4,7 @@ use std::io;
 use std::io::Write;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
-use std::sync::{PoisonError, RwLock};
+use std::sync::{Mutex, PoisonError, RwLock};
 
 use indexmap::IndexMap;
 use metrics::{Counter, Gauge, Histogram, Key, KeyName, Metadata, Recorder, SharedString, Unit};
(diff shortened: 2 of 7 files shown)


metrics-exporter-prometheus/Cargo.toml
```diff
@@ -65,6 +65,7 @@ tracing = { workspace = true, optional = true }
 mime = { version = "0.3", optional = true }
 prost = { workspace = true, optional = true }
 prost-types = { workspace = true, optional = true }
+evmap = "11.0.0"
```
tobz: Can we move this to be defined in the workspace `Cargo.toml`, following the pattern of the other dependencies?

metrics-exporter-prometheus/src/recorder.rs
```diff
@@ -336,15 +379,21 @@ impl Recorder for PrometheusRecorder {
     }
 
     fn register_counter(&self, key: &Key, _metadata: &Metadata<'_>) -> Counter {
-        self.inner.registry.get_or_create_counter(key, |c| c.clone().into())
+        let counter = self.inner.registry.get_or_create_counter(key, |c| c.clone().into());
+        self.inner.commit_outstanding_description_writes();
```
tobz: Why are we also committing outstanding updates here, during the register call and not during the `describe_*` methods?

tobz (changes requested):

tobz:
> Question for maintainers: If this approach makes sense, I am a little unsure on whether the commit points I've added are sufficient for typical uses. I'm open to discussion on what makes sense.

I'm don't believe I understand the question here.

I'm on board with the description of the problem, insofar as I agree that it sounds like a legitimate problem. I'm on board with trying to use `evmap` to solve that problem: I don't have enough context to say like "yeah, 100%, this is the best way forward and has no pitfalls" but in general, the approach here doesn't _concern_ me or fire off any alarms in my head of being _obviously_ worse for some other use cases, etc.

Does that answer your question?

tobz (approved):

tobz:

PR #681, opened 2026-03-03
feat(metrics-util): Expose true count and sum on Drain for reservoir sampling

michelhe:
## Summary

Expose two public methods on `Drain` that let downstream consumers recover accurate aggregate statistics when reservoir sampling drops values:

- **`Drain::unsampled_len()`** — returns the total number of samples pushed into the reservoir, including those dropped by sampling. The field was already tracked internally; this just adds a public getter.
- **`Drain::unsampled_sum()`** — returns the true sum of all pushed samples, including dropped ones. Atomically accumulated in `Reservoir::push()` using a `fetch_update` CAS loop on `AtomicU64` (same pattern as `GaugeFn::increment` in `metrics/src/atomics.rs`). Reset in `Drain::drop()` alongside the existing count reset.

## Motivation

When the reservoir overflows, downstream consumers draining the samples only see the sampled subset. A consumer naively counting or summing values in the drain loop would get results capped at reservoir capacity — not the true totals. For example, pushing 1000 values into a 16-slot reservoir means the drain yields only 16 samples, so a naive count gives 16 and a naive sum covers only those 16 retained values, while the true count is 1000 and the true sum is 499,500.

Previously, the only way to recover the true count was `drained_count / sample_rate`, which is a lossy float round-trip. There was no way to recover the true sum at all.

## Tests

Unit tests covering: under-capacity (no sampling), overflow (proving true values survive), reset-after-drain (A/B buffer reuse), and empty reservoir.

### DIFF
--- a/metrics-util/src/storage/reservoir.rs
+++ b/metrics-util/src/storage/reservoir.rs
@@ -214,3 +214,78 @@ impl AtomicSamplingReservoir {
         f(drain);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_under_capacity_no_overflow() {
+        let reservoir = AtomicSamplingReservoir::new(64);
+
+        for i in 0..10 {
+            reservoir.push(i as f64);
+        }
+
+        reservoir.consume(|drain| {
+            assert_eq!(drain.unsampled_len(), 10);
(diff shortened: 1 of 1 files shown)


tobz (approved):
Looks good to me. 👍🏻

tobz:
Released as `metrics-util@v0.20.2`.

PR #686, opened 2026-04-01
feat(prometheus): refactor exporter to yield structured data

demosdemon:
My core need was the ability to export metrics over an FFI layer without intermediate encoding. See https://github.com/ava-labs/firewood/pull/1868 for this in action.

This pull request introduces a new **structured rendering API** for Prometheus metrics and refactors the **protobuf** renderer to sit on top of it. The text exposition format is unchanged.

## Rendering and API improvements

- Added a new public `render` module that provides a format-agnostic, structured representation of a metrics snapshot (`MetricFamily`, `Metric`, `LabelPair`, `MetricValue`, and the typed value structs), exposed as the `RenderedMetrics` iterator.
- Each `MetricFamily` carries an explicit `kind` (`MetricKind`) and the registered `unit`, so consumers get the full Prometheus data model without an intermediate encoding.
- Introduced `PrometheusHandle::render_snapshot_and_descriptions` (and the corresponding `Inner` method) to obtain this structured data directly. The `render` module is always compiled, so this API is available without the `protobuf` feature.

## Protobuf rendering refactor

- `render_protobuf` and `render_protobuf_to_write` now consume `RenderedMetrics` and convert it to the protobuf wire format via a thin `into_protobuf()` layer, replacing the previous monolithic renderer.
- The protobuf **wire output is unchanged**: a golden-byte regression test renders a representative snapshot and asserts the bytes are identical to those produced by `main`.

## Code cleanup

- Removed `#[cfg(...)]` feature gates from several `NativeHistogram` accessor methods, which are now used unconditionally by the always-compiled `render` module.
- Made the native-histogram bucket reads tolerate a poisoned lock, consistent with the rest of the recorder.

## Tests

- Added unit tests for the native-histogram span/delta encoder (`make_buckets`), the `RenderedMetrics` iterator contract (ordering + `ExactSizeIterator`), distribution rendering (classic histogram, native histogram, summary) through protobuf, and the structured `kind`/`unit` fields.


tobz:
👋🏻 

I started taking a look at this but got a little sidetracked. I plan to take another look in the coming days.

rkuris:
Hi @tobz! Just checking in on this PR. We're using this fork in the Firewood 
project (https://github.com/ava-labs/firewood) for FFI metrics export and it's 
working well. The fork is now 32 commits behind main, so we'd love to see this 
merged upstream. Happy to rebase or address any concerns!

PR #688, opened 2026-04-16
feat: make key retention more efficient

nmooscisco:
In the export paths for the `metrics-exporter-*` crates present in this repository, the `metrics::key::Key` type necessitates a lot of cloning, which for high-cardinality series, can be quite expensive. This patch aims to reduce that cost by introducing a `retained` mode for keys and their component parts: names and label arrays.

This change has the following key features:

1. `Cow<'static, [T]>` (used for labels) and `Cow<'static, str>` (used for names) have been given a new method `to_retained`, which either clones the contents into a `Shared` variant, or clones the existing `Cow`, preserving its `Shared` or `Borrowed` (static) state.
2. Introduced `KeyName::to_retained` and `Key::to_retained` based on the above new methods, which is intended to be used to save the object for storage in a registry or other long-lived data structure.
3. The metrics exporters that make use of a registry in this crate have been modified to convert the key into a retained key at the boundary of their instantiation.

This will significantly reduce the need for allocations when exporting a large number of metrics, as keys stored in this new retained mode will be efficient to clone.

### DIFF
--- a/metrics/src/key.rs
+++ b/metrics/src/key.rs
@@ -28,6 +28,15 @@ impl KeyName {
     pub fn into_inner(self) -> SharedString {
         self.0
     }
+
+    /// Returns a version of this name that is cheap to clone for long-lived retention.
+    ///
+    /// *NOTE:* This will allocate if this `KeyName` was created from a non-`'static`
+    /// string, however, the returned `KeyName` will not require allocation when cloned
+    /// or `to_retained` is invoked again.
+    pub fn to_retained(&self) -> Self {
+        KeyName(self.0.to_retained())
+    }
 }
 
 impl<T> From<T> for KeyName
@@ -143,6 +152,15 @@ impl Key {
     /// Consumes this [`Key`], returning the name parts and any labels.
     pub fn into_parts(self) -> (KeyName, Vec<Label>) {
         (self.name, self.labels.into_owned())
+    }
+
+    /// Returns a version of this key that is cheap to clone for long-lived retention.
+    ///
+    /// *NOTE:* This will allocate if this `Key` was created from non-`'static`
+    /// parts, however, the returned `Key` will not require allocation when cloned
+    /// or `to_retained` is invoked again.
+    pub fn to_retained(&self) -> Self {
+        Self { name: self.name.to_retained(), labels: self.labels.to_retained(), hash: self.hash }
     }
 
     /// Clones this [`Key`], and expands the existing set of labels.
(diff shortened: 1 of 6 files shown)


metrics/src/key.rs
```diff
@@ -28,6 +28,11 @@ impl KeyName {
     pub fn into_inner(self) -> SharedString {
         self.0
     }
+
+    /// Returns a version of this name that is cheap to clone for long-lived retention.
```
tobz, line 32: We should augment this comment to mention that if the key was constructed from non-`'static` components, it will allocate when creating this new `Key` but _not_ when that `Key` is subsequently cloned or `to_retained` is called again.

metrics/src/key.rs
```diff
@@ -145,6 +150,11 @@ impl Key {
         (self.name, self.labels.into_owned())
     }
 
+    /// Returns a version of this key that is cheap to clone for long-lived retention.
```
tobz, line 157: Same comment here as the other comment on `to_retained`.

tobz (changes requested):
Left two small comment nits, but I think the bigger thing is: maybe we should actually just do the `Key::to_retained` call in the registry itself... then users can't forget to take advantage of it.

metrics/src/key.rs
nmooscisco: Expanded the documentation comment here.

metrics/src/key.rs
nmooscisco: Same here.

nmooscisco:
The problem with doing the `to_retained` invocation inside the registry itself is that the registry is generic over `K`, not specifically requiring a `Key`. There _is_ a way to do this, but it involves creating either 1) a post-processing function for the key that is bound to the generic `K`, or 2) an additional trait to bound `K` with. I can attempt those if you would like -- I do like the idea of the registry itself being responsible for properly retaining the objects it's maintaining.

nmooscisco:
I prototyped a concept for moving the `to_retained` invocation into the registry without a semver compatibility break [1]. Let me know what you think.

If you're okay with a semver compatibility break, the change can be smaller at the expense of requiring the downstreams to also be updated.

I opened a second PR because I didn't want to clutter this one with too many changes. Will close that one if you would prefer a different approach.

[1]: https://github.com/metrics-rs/metrics/pull/689/changes/1b0de18f3da2e982240acafec0004d20233eb09e

tobz:
> The problem with doing the `to_retained` invocation inside the registry itself is that the registry is generic over `K`, not specifically requiring a `Key`. There _is_ a way to do this, but it involves creating either 1) a post-processing function for the key that is bound to the generic `K`, or 2) an additional trait to bound `K` with. I can attempt those if you would like -- I do like the idea of the registry itself being responsible for properly retaining the objects it's maintaining.

Oh, duh, right.

Let's just leave it as-is for now then. Can always generalize later.

tobz (approved):

tobz:
Apologies on the really delayed review here. Life has been busy. 😅 

LGTM!

tobz:
Released as `metrics@v0.24.4`, `metrics-util@v0.20.2`, `metrics-exporter-prometheus@v0.18.2`, and `metrics-exporter-dogstatsd@v0.9.7`.

PR #690, opened 2026-04-28
fix(perf): ensure hash_label properly hashes both the key and value

nmooscisco:
Previously, the `hash_label` function unintentionally hashes the label key twice, which causes all metrics with the same name and set of label names to be hashed to the same bucket, causing quadratic-time hash table insertion performance when building snapshots on rendering. This patch fixes the bug and defensively adjusts the `hash_label` function to ensure this kind of problem is not possible.


tobz (approved):

tobz:
Released as `metrics@v0.24.5`.

PR #691, opened 2026-04-29
enhancement(metrics): allow specifying description and unit in registration macros

DoumanAsh:
When I initially integrated metrics crate into my code I had one inconvenience that I needed to describe and create counter separately, even though I technically only create it once...

Putting aside the fact that you need to always describe metric before creating it, I always end up creating macro that would wrap description and creation into one beautiful macro
So I want to propose such interface here

The basic idea is to use optional labels to `describe` which in turn allows to add optional tag `unit`
Then in order not to confuse macro parser (due to presence of `describe`), I wrapped labels into optional brackets `{ label1 => value2 }`

This will allow user to describe macro in one go with creation
But this is admittedly very complex, so let me know if you're interested in this approach and I'll add macros for other metrics

### DIFF
--- a/metrics/src/macros.rs
+++ b/metrics/src/macros.rs
@@ -297,79 +297,6 @@ macro_rules! __register_metric {
     }};
 }
 
-/// Registers a counter with provided description
-///
-///## Usage
-///
-/// ```rust
-/// # #![no_implicit_prelude]
-/// # use ::std::convert::From;
-/// # use ::std::format;
-/// # use ::std::string::String;
-/// # use metrics::create_counter as create_counter;
-/// # fn main() {
-/// // A basic counter:
-/// let counter = create_counter!("some_metric_name");
-/// counter.increment(1);
-///
-/// // A basic counter with labels
-/// let counter = create_counter!("some_metric_name", "label1" => "value2" );
-/// counter.increment(1);
-///
-/// // A counter with description!
-/// let counter = create_counter!(describe: "my super counter", "some_metric_name");
-/// counter.increment(1);
-///
-/// // A counter with description and unit!
-/// let counter = create_counter!(describe: "my super counter", unit: metrics::Unit::Bytes, "some_metric_name");
-/// counter.increment(1);
-///
-/// // A custom level counter with description and unit!
-/// let counter = create_counter!(
-///     describe: "my super counter",
-///     unit: metrics::Unit::Bytes,
-///     level: metrics::Level::INFO,
-///     "some_metric_name",
-/// );
-/// counter.increment(1);
-///
-/// // A custom target counter with description and unit!
-/// let counter = create_counter!(
-///     describe: "my super counter",
-///     unit: metrics::Unit::Bytes,
-///     target: "custom_target",
-///     "some_metric_name",
-/// );
-/// counter.increment(1);
-///
-/// // A counter with description and fancy label, even target module!
-/// let counter = create_counter!(
-///     describe: "my super counter",
-///     unit: metrics::Unit::Bytes,
-///     target: ::core::module_path!(),
-///     level: metrics::Level::INFO,
-///     "some_metric_name",
-///     "label1" => "value1",
-///     "label2" => "value2"
-/// );
-/// counter.increment(1);
-/// # }
-/// ```
-#[macro_export]
-macro_rules! create_counter {
-    ($($input:tt)*) => {
-        $crate::__register_metric!(
-            describe_counter,
-            register_counter,
-            describe = __internal_metric_description_none__,
-            unit = __internal_metric_unit_none__,
-            target = ::core::module_path!(),
-            level = $crate::Level::INFO;
-            $($input)*
-        )
-    };
-}
-
 /// Registers a gauge.
 ///
 /// Gauges represent a single value that can go up or down over time, and always starts out with an
@@ -446,9 +373,8 @@ macro_rules! create_counter {
 /// let gauge = gauge!(name);
 ///
 /// let gauge = gauge!(format!("{}_via_format", "name"));
-///
 /// //Full gauge customization example
-/// let counter = gauge!(
+/// let gauge = gauge!(
 ///     describe: "super gauge",
 ///     unit: metrics::Unit::Bytes,
 ///     target: ::core::module_path!(),
@@ -548,7 +474,7 @@ macro_rules! gauge {
 ///
 /// let histogram = histogram!(format!("{}_via_format", "name"));
 /// //Full histogram customization example
-/// let counter = histogram!(
+/// let histogram = histogram!(
 ///     describe: "super counter",
 ///     unit: metrics::Unit::Bytes,
 ///     target: ::core::module_path!(),
--- a/metrics/src/macros.rs
+++ b/metrics/src/macros.rs
@@ -68,14 +68,37 @@ macro_rules! key_var {
 /// which includes using macros such as `format!` directly at the callsite. String literals are
 /// preferred for performance where possible.
 ///
+/// # Usage
+///
+/// ```
+/// [opt: value] <$name> [$labels]
+/// ```
+///
+/// Only `$name` is required to initialize metrics.
+///
+/// All `opt`s MUST be specified before `$name` while `$labels` parameter block always go after `$name`
+///
+/// Following is brief explanation of parameters
+///
+/// ## Required parameters
+///
+/// - `$name` - Name of the metric. Can be expression that results in `String` or `&'static str`
+///
 /// ## Optional Parameters
 ///
 /// Following parameters can be provided in any order
 ///
-/// `target:` - Specifies counter target. Defaults to `::core::module_path!()`.
-/// `level:` - Specifies counter level. Defaults to `INFO`.
-/// `describe:` - Specifies counter description to register for counter. If specified `$name` will be used twice.
-/// `unit:` - Specifies counter unit to register for counter if `describe:` is specified.
+/// - `target:` - Specifies counter target. Defaults to `::core::module_path!()`.
+/// - `level:` - Specifies counter level. Defaults to `INFO`.
+/// - `describe:` - Specifies counter description to register for counter. If specified `$name` will be used twice.
+/// - `unit:` - Specifies counter unit to register for counter if `describe:` is specified.
+///
+/// ## Labels
+///
+/// Labels can be passed as _one_ of following:
+/// - Arbitrary number of `<key> => <value>` where `key` and `value` can be expression that results in `&'static str` or `String`
+/// - Static reference to collection of **Label**
+/// - Collection/iterator that implements [IntoLabels](trait.IntoLabels.html)
 ///
 /// # Example
 /// ```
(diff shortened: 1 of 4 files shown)


tobz:
👋🏻 

Thanks for submitting this PR! I agree that the two separate macros are an annoyance/sharp edge in terms of the API and so I'm interested in approaches for improving this situation.

Looking at the example usages, I don't love the need to change how labels are passed in when providing the new labeled parameters. My preference would be to keep the current structure where labels can either be a convertible expression or `key => value` pairs.

I wonder if we could somehow use a TT muncher approach to incrementally handle any number of named parameters at the _front_ of the macro, and in any order, while still matching against the required name/name-and-labels portions at the end.

DoumanAsh:
> Looking at the example usages, I don't love the need to change how labels are passed in when providing the new labeled parameters. My preference would be to keep the current structure where labels can either be a convertible expression or key => value pairs.

Got it, let me think if I can avoid need for brackets while enabling description to be optional

DoumanAsh:
@tobz I did few iterations, but at first glance I cannot make it very well right now
https://github.com/metrics-rs/metrics/pull/691/changes/e459e2618b89f393059b7dcfc79af160e66541f4

I will think on it more, but current version allows to be drop in replacement of `counter!`
But I could simplify it and only leave branches for handling `description + unit` and then pass everything to `counter!`

I do not think it would be possible to break down macro parser because both `counter` and `describe` macros need `$name` so one way or another I would have to describe all possible paths that cover minimal necessary arguments alongside `$name`

But I will give it some thinking today and see if I can simplify it further.

DoumanAsh:
@tobz I was able to do small simplification by making target/level optional via fallback macro
https://github.com/metrics-rs/metrics/pull/691/changes/4422c162ac77f292d43d9b7b5d0f86e199f56b2d

This allows to remove extra branch and relax order of target/level in addition to description/unit

It is still not exactly looking slim, but I'm not sure if it is possible to do better with constraints of existing API

tobz:
@DoumanAsh I was inspired by your experimentation to keep trying to figure out how to make a TT muncher-based design work for us, and I came up with this: https://github.com/metrics-rs/metrics/compare/main...tobz/tt-muncher-macros

The trick was having an inner macro that each leaf macro (`counter!`, etc) calls out to, where it defines named parameter accumulators ahead of the muncher. The accumulators are always in a fixed position, and leaf macros always call the inner macro with those accumulators set to a default value... so we munch named parameters until none are left, and then we just have `<accumulator tokens>, <normal stuff>`, where `<normal stuff>` matches the metric name and labels, if present.

Right now, that branch only implements `target` and `level`, but I think it can be extended to add `description` and `unit` if we treat those as being `Option` and then, in the "terminator" pattern, we generate the code to actually say something like:

```
if description is set || unit is set {
   describe_<metric type>!(description, unit)
}

// register metric
```

We would set those accumulators to default to `None`, so that it didn't fire when they weren't set, but if either was set, we do the describe call before initially registering the metric.

DoumanAsh:
I see, yeah, that certainly could be one way
I didn't like this approach because it would mean much more extra recursion , and it is quite easy to misuse as you can repeat macro forever, even if normal use case should not cause such issues.

I will update my PR to follow this approach and ping you to review it once I handle both description and target/level

DoumanAsh:
@tobz Please find my idea how to make optional fields for description
https://github.com/metrics-rs/metrics/pull/691/changes/a9caa57190d736f4f7b2553a7e582cc9a03395c3
I expanded your idea fully to avoid need for embedding IFs, instead it just uses special internal idents to dispatch description creation based on whether user provided input or not
Let me know if it is ok and I will proceed to create macros and documentation for all metrics (or replace existing macroses)

P.s. error from incorrect usage will not look nice to user, but I think it can be improved by adding comment to every branch to clarify the reason for failing to match

metrics/src/macros.rs
```diff
@@ -115,10 +137,352 @@ macro_rules! counter {
...
+    (
+        $describe:ident,
+        $register:ident,
+        describe = $description:tt,
+        unit = $unit:tt,
+        target = $target:expr,
+        level = $level:expr;
+        $name:expr $(, $label_key:expr $(=> $label_value:expr)?)* $(,)?
+    ) => {{
+        $crate::__describe_metric!(describe_counter, $description, $unit, $name);
```
DoumanAsh: TODO: need to document that `$name` will be used twice in case `describe:` is specified

tobz:
> @tobz Please find my idea how to make optional fields for description [a9caa57](https://github.com/metrics-rs/metrics/commit/a9caa57190d736f4f7b2553a7e582cc9a03395c3) I expanded your idea fully to avoid need for embedding IFs, instead it just uses special internal idents to dispatch description creation based on whether user provided input or not Let me know if it is ok and I will proceed to create macros and documentation for all metrics (or replace existing macroses)

@DoumanAsh Yeah, this is looking better. 👍🏻 

I think if I'm understanding right... `internal_target_fallback`, `internal_level_fallback`, and `create_counter` are leftover code and `__describe_metric`, `__register_metric`, and `create_counter` are all the cleaned up approach?

If I have that right, then yeah, I think we'd want to go with the cleaned up approach, updating the existing macros (`counter!`, etc) to use it. I believe we also need a fourth match arm on `__describe_metric!` that handles the case when the description is unset but unit is set, since that's not currently possible with `Recorder::describe_*`.

DoumanAsh:
> I think if I'm understanding right... internal_target_fallback, internal_level_fallback, and create_counter are leftover code and __describe_metric, __register_metric, and create_counter are all the cleaned up approach?

Yes, that's correct

> I believe we also need a fourth match arm on __describe_metric! that handles the case when the description is unset but unit is set, since that's not currently possible with Recorder::describe_*

Yeah, that's right, I forgot about this case
It should be compile error currently since we cannot really use `unit` without description so I will add this branch to `__describe_metric` macro

DoumanAsh:
I updated all metrics macroses to use new internal macro and added test file

tobz:
👍🏻 I'll try to sit down and do a final review in the next few days.

metrics/src/macros.rs
```diff
@@ -68,6 +68,36 @@ macro_rules! key_var {
...
+///
+/// - `$name` - Name of the metric. Can be expression that results in `String` or `&'static str`
+///
+/// ## Optional Parameters
+///
+/// Following parameters can be provided in any order
+///
+/// - `target:` - Specifies counter target. Defaults to `::core::module_path!()`.
+/// - `level:` - Specifies counter level. Defaults to `INFO`.
+/// - `describe:` - Specifies counter description to register for counter. If specified `$name` will be used twice.
```
tobz: Is the "will be used twice" comment here referring to the fact we'll end up evaluating the expression given for `name` two times: once for the describe call, once for the registration?

metrics/src/macros.rs
```diff
@@ -100,26 +130,171 @@ macro_rules! key_var {
 /// let name = String::from("some_owned_metric_name");
 /// let counter = counter!(name);
 ///
-/// let counter = counter!(format!("{}_via_format", "name"));
+/// let gauge = counter!(format!("{}_via_format", "name"));
+///
+/// //Full counter customization example
```
tobz:
```suggestion
/// // Full counter customization example
```

(and apply the same change to the equivalent docs example of the other macros)

metrics/src/macros.rs
```diff
@@ -68,6 +68,36 @@ macro_rules! key_var {
...
+///
+/// Only `$name` is required to initialize metrics.
+///
+/// All `opt`s MUST be specified before `$name` while `$labels` parameter block always go after `$name`
+///
+/// Following is brief explanation of parameters
+///
+/// ## Required parameters
+///
+/// - `$name` - Name of the metric. Can be expression that results in `String` or `&'static str`
```
tobz:
```suggestion
/// - `$name` - Name of the metric. Must be a string literal or an expression that results in `String` or `&'static str`.
```

(and apply the same change to the docs for the other macros)

metrics/src/macros.rs
```diff
@@ -68,6 +68,36 @@ macro_rules! key_var {
...
+///
+/// ## Optional Parameters
+///
+/// Following parameters can be provided in any order
```
tobz:
```suggestion
/// The following parameters can be provided in any order:
```

metrics/src/macros.rs
```diff
@@ -68,6 +68,36 @@ macro_rules! key_var {
...
+///
+/// - `target:` - Specifies counter target. Defaults to `::core::module_path!()`.
+/// - `level:` - Specifies counter level. Defaults to `INFO`.
+/// - `describe:` - Specifies counter description to register for counter. If specified `$name` will be used twice.
+/// - `unit:` - Specifies counter unit to register for counter if `describe:` is specified.
```
tobz:
```suggestion
/// - `target:` - Module path of the counter. Defaults to `::core::module_path!()`.
/// - `level:` - Verbosity level of the counter. Defaults to `INFO`.
/// - `description:` - Description of the counter. If specified, `$name` will be used twice. 
/// - `unit:` - Unit of measurement of the counter. Description must be provided in order to specify units.
```

(apply the same change to docs for the other macros)

metrics/src/macros.rs
```diff
@@ -68,6 +68,36 @@ macro_rules! key_var {
...
+///
+/// ## Labels
+///
+/// Labels can be passed as _one_ of following:
+/// - Arbitrary number of `<key> => <value>` where `key` and `value` can be expression that results in `&'static str` or `String`
```
tobz:
```suggestion
/// - Arbitrary number of `<key> => <value>` where `key` and `value` can be a string literal or an expression that results in `String` or `&'static str`.
```

(and apply this same change to the docs for other macros)

metrics/src/macros.rs
```diff
@@ -100,26 +130,171 @@ macro_rules! key_var {
 /// let name = String::from("some_owned_metric_name");
 /// let counter = counter!(name);
 ///
-/// let counter = counter!(format!("{}_via_format", "name"));
+/// let gauge = counter!(format!("{}_via_format", "name"));
```
tobz: Seems like `counter` got mistakenly changed to `gauge` here?

tobz (changes requested):
The macro changes themselves look almost complete to me, plus some documentation cleanup before it's ready.

metrics/src/macros.rs
tobz: Also, forgot to mention this as a separate review comment but we should change `describe` to `description`.

metrics/src/macros.rs
DoumanAsh: Yes, hence user needs to be aware if `$name` computation is heavy or it is only name of variable (in case of variable you cannot pass String for example as it will be used more than once)

DoumanAsh:
@tobz Thanks for review, I applied comments and renamed `describe` parameter

tobz (approved):

tobz:
Released as `metrics@v0.24.6`.

Thanks for your contribution! :heart:

PR #696, opened 2026-06-13
Change from_arc to allow existing trait object

DoumanAsh:
Current ctor for taking existing Arc objects require concrete type and will not allow existing `dyn TraitFn + Send + Sync + 'static`

This prevents user from passing their CounterFn etc as trait object already wrapped in Arc making it impossible to dynamically create counters for implementors.

For example opentelemetry has 3 ways to implement counters which I want to use, but it is impossible for me to store counters as `Arc<dyn CounterFn>` because `metrics::Counter::from_arc` will not allow to create counter from Arc containing `!Sized` (e.g. trait object)

~This change should be backward compatible because Arc with concrete type always coerces into trait object as long as there is no extra layers between pointers~

This change is not backward compatible since it will not play well with `Into` and `Arc::clone`

### DIFF
--- a/metrics-exporter-dogstatsd/src/recorder.rs
+++ b/metrics-exporter-dogstatsd/src/recorder.rs
@@ -24,20 +24,20 @@ impl Recorder for DogStatsDRecorder {
         let key = key.to_retained();
         self.state
             .registry()
-            .get_or_create_counter(&key, |existing| Counter::from_arc(Arc::clone(existing)))
+            .get_or_create_counter(&key, |existing| Counter::from_arc(existing.clone()))
     }
 
     fn register_gauge(&self, key: &Key, _: &Metadata<'_>) -> Gauge {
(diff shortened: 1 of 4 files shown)


tobz:
Hmm, interesting.

Question I genuinely don't the answer to: could we just add `?Sized` to the generic bound instead?

DoumanAsh:
Not really, you have to pass `dyn T` directly if you want unsized object, otherwise with `?Sized` you're unable to coerce into `dyn T`
Rust requires concrete type for trait object conversion (it would work with super trait coercing though)

Feel free to keep it for next major bump
I found work around for my issue by just using metrics' primitives in my cache instead of OTEL's one

Reference: https://doc.rust-lang.org/reference/type-coercions.html#r-coerce.unsize.trait-object
https://doc.rust-lang.org/std/marker/trait.Unsize.html

PR #698, opened 2026-07-14
Make Histogram::record_many O(1) end to end

rpb-ant:
## Motivation

`record_many(value, count)` is documented as recording a value `count` times, but every implementation falls back to the trait's default loop of `count` `record()` calls. In `metrics-exporter-prometheus`, each call buffers a `(value, Instant)` sample until the next upkeep, so `record_many(v, 10_000_000)` buffers tens of millions of samples. This makes `record_many` unusable as a weighted observe (e.g. occupation-time histograms recording "N microseconds at level L" in one call).

## Change

`record_many` becomes O(1) in `count` on every path:

- `metrics`: `HistogramFn for Arc<T>` forwards `record_many` (wrapper chains previously fell back to the default loop).
- `metrics-util`: `Generational<T>` and `FanoutHistogram` forward `record_many`; new `Histogram::record_n` and `Summary::add_n`.
- `metrics-exporter-prometheus`: the registry buffers one `(value, count, timestamp)` entry per call; all three distribution types apply counts in constant time on drain — bucketed histograms via `record_n`, summaries via `RollingSummary::add_n` → `Summary::add_n` → `DDSketch::add_with_count`, native histograms via `observe_n`.

Summaries are the default distribution type, so an O(count) summary drain would leave the failure mode in place on exactly the unconfigured path.

Scope: O(1) applies to the prometheus exporter path. Impls whose contract is raw-sample storage (`AtomicBucket<f64>`, the sampling reservoir, dogstatsd/tcp exporters) keep the default loop — unchanged behavior.

## Breaking (exporter)

`Distribution::record_samples` now takes `&[(f64, usize, Instant)]` (`RollingSummary::count` also widens to `u64`), and `AtomicBucketInstant::clear_with` callbacks receive counted entries. Cost note: the buffered entry widens from 16 to 24 bytes, +50% inter-upkeep buffer for plain `record()` users; there is no packing that avoids it while keeping per-sample timestamps.

## Blocked on

mheffner/rust-sketches-ddsketch#22 (`DDSketch::add_with_count`) shipping in a release; the workspace dep is bumped to `0.4` in anticipation. Until then, the workspace builds and tests with a `[patch.crates-io]` path override.

## Tests

Forwarding tests at each wrapper layer (a mock distinguishes forwarding from the default loop); equivalence tests (`record_n`/`add_n`/`observe_n` ≡ repeated single records, including NaN and zero-count edges); a regression test asserting one `record_many` call buffers exactly one entry; end-to-end render tests for the classic-bucket and summary paths.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ rand_xoshiro = { version = "0.7", default-features = false }
 rapidhash = { version = "4.4.1", default-features = false }
 ratatui = { version = "0.29", default-features = false }
 rustls = { version = "0.23", default-features = false }
-sketches-ddsketch = { version = "0.3", default-features = false }
+sketches-ddsketch = { version = "0.4", default-features = false }
 thiserror = { version = "2", default-features = false }
 tokio = { version = "1", default-features = false, features = ["rt", "net", "time", "rt-multi-thread"] }
 tracing = { version = "0.1", default-features = false }
--- a/metrics-exporter-prometheus/CHANGELOG.md
+++ b/metrics-exporter-prometheus/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- **Breaking:** `Distribution::record_samples` now takes `(f64, usize, Instant)` entries — the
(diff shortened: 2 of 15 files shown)


rpb-ant:
Hrm...I just did some actual analysis about how much the extra `count` would cost in the buffer and it's non-trivial. Closing this out while I consider alternatives.

PR #701, opened 2026-08-02
feat(metrics-exporter-datadog): sanitize metric name and labels by default

### LINKED ISSUE #603: Sanitize metric labels in `metrics-exporter-dogstatsd` to avoid generating invalid DogStatsD payloads.
## Context

DogStatsD as a protocol is underspecified in terms of what it will accept which can lead to issues where a label value is theoretically valid but generate an _invalid_ DogStatsD payload. A simple example of this is a label where the value contains the pipe character: `|`. This character is essentially reserved for separating structured fields in the DSD payload.

The [public Datadog documentation for tagging](https://docs.datadoghq.com/getting_started/tagging/#define-tags) states that tags will be sanitized according to a small ruleset. When sending the same type of labels as described above, the values would be sanitized such that they would actually allow generating a valid DSD payload. Following the existing example, if our label value was `Foo|Bar`, it would be sanitized to `foo_bar` on the Datadog platform. If we simply sent `foo_bar` in the first place, then the DSD payload would be valid.

We should add an enabled-by-default sanitization step to labels so that we generate them in a way that matches the Datadog tagging rules and allows for them to generate valid DSD payloads. This should be enabled by default because the exporter targets communication with the Datadog Agent. We should add a way to opt-out of this behavior if users are targeting something _else_.

Guflly:
Sanitize DogStatsD labels by default before writing payloads.

The sanitizer lowercases tags, replaces invalid runs, trims underscores, and enforces the 200-character limit. `with_label_sanitization(false)` keeps raw labels for non-Datadog targets.

Fixes #603

Tests:
- `cargo test -p metrics-exporter-dogstatsd --all-features`
- `cargo clippy -p metrics-exporter-dogstatsd --all-features --all-targets`
- `cargo fmt --all -- --check`

### DIFF
--- a/metrics-exporter-dogstatsd/src/builder.rs
+++ b/metrics-exporter-dogstatsd/src/builder.rs
@@ -98,6 +98,7 @@ pub struct DogStatsDBuilder {
     histogram_sampling: bool,
     histogram_reservoir_size: usize,
     histograms_as_distributions: bool,
+    sanitize_labels: bool,
     global_labels: Vec<Label>,
     global_prefix: Option<String>,
 }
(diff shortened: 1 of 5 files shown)


tobz:
👋🏻 

I ended up push a commit to fix some issues with sanitization of labels, and also to just go ahead and add sanitization of metric names as well.

PR #702, opened 2026-08-03
fix(metrics-exporter-prometheus): use portable atomics

madadam:
Use `metrics::atomics::AtomicU64` instead of `std::sync::atomic::AtomicU64` so the crate builds for platforms lacking 64-bit atomics support.


tobz (approved):

tobz:
Just one small formatting nit, but should be good to go after that. 👍🏻

PR #704, opened 2026-08-19
fix(metrics-exporter-prometheus): don't evict metrics with live handles

michelhe:
<NOTE: This is marked as draft as I need to self-review more thoroughly>

Callers may hold a Counter/Gauge/Histogram handle for the lifetime of the process to keep registration off the hot path. With idle_timeout set, the exporter deleted the registry entry for such a metric once it went idle, which fails at both things eviction is for: the storage isn't reclaimed, since the caller's Arc keeps it alive, and if that handle is the metric's only producer, nothing ever re-registers it, so its later updates land in storage the registry can no longer reach. The metric goes silent for the rest of the process's life.

Recency now sweeps the registry through retain_*, which lends it the map's own copy of each handle: a reference count above one means the application is still holding that metric, and the entry stays. Deciding under the shard lock also means a handle handed out by a concurrent registration cannot be orphaned. A metric that was held is evicted by the first sweep after its last handle is dropped, without a fresh idle period.

This replaces Recency::should_store_* with Recency::evict_idle_*, which is a breaking change for metrics-util.

### DIFF
--- a/metrics-exporter-prometheus/CHANGELOG.md
+++ b/metrics-exporter-prometheus/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- `idle_timeout` no longer removes metrics whose handle is still held by the application. Removing
+  them freed nothing, since the handle owns the metric's storage, and it stranded the holder: its
+  later updates landed in storage no longer reachable from the registry, so the metric stayed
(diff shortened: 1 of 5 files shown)


michelhe:
So far testing looks good but I think the code can be de-SLOPed

tobz:
I haven't even looked at the code here yet, to be completely honest... but I think whatever breaking changes are needed to make it so that handle liveness is the one and only mechanism that controls whether a metric stays alive or not is worth exploring here.

This is probably the biggest outstanding problem with the design as-is, and it's not unreasonable to fix it via a breaking change if that's what it takes.

PR #705, opened 2026-08-24
fix(metrics-exporter-dogstatsd): support IPv6 remote addresses

funcpp:
`metrics-exporter-dogstatsd` can't send to an IPv6 remote address. In `forwarder/sync.rs` the UDP socket is always bound to `Ipv4Addr::UNSPECIFIED` regardless of the remote's address family, so connecting the resulting `AF_INET` socket to an IPv6 remote is rejected by the kernel (`EAFNOSUPPORT` on Linux, `EINVAL` on macOS).

The failure is silent. `RemoteAddr::try_from` uses `to_socket_addrs()`, so `[::1]:8125` and `udp://[fd00::1]:8125` parse fine and neither `with_remote_address()` nor `build()` returns an error. Socket creation is deferred to the forwarder thread's first flush, and on failure the state machine falls back to `Disconnected` and retries on the next one. So the application starts up normally and drops every metric, logging only `Failed to send payload.` once per flush interval.

This patch selects the bind address family from the first remote address, with a unit test covering that selection. I also verified locally that a datagram now reaches a real `[::1]` listener, but left that round trip out as a test since it would be flaky on runners without IPv6 loopback.

Known limitation: `connect(&addrs[..])` walks the whole address list, but the socket family is fixed by the first entry, so a mixed-family list can still fail on the rest. Doing that properly needs a socket per attempt inside the connect loop, which felt like more surgery than this bug warrants — happy to fold it in if you'd rather. It does mean a hostname resolving AAAA-first now connects over IPv6 instead of falling through to IPv4; the default `127.0.0.1:8125` is unaffected.

### DIFF
--- a/metrics-exporter-dogstatsd/src/forwarder/sync.rs
+++ b/metrics-exporter-dogstatsd/src/forwarder/sync.rs
@@ -2,7 +2,7 @@
 use std::os::unix::net::{UnixDatagram, UnixStream};
 use std::{
     io::{self, Write as _},
-    net::{Ipv4Addr, UdpSocket},
+    net::{Ipv4Addr, Ipv6Addr, SocketAddr, UdpSocket},
     sync::Arc,
     thread::sleep,
     time::Instant,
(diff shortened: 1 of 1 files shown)


tobz:
I ended up making a small change to this PR to prefer IPv4 over IPv6 when _both_ are present in the address list. This aligns with how the Datadog Agent actually binds its UDP listener socket which is that unless `bind_host` is set explicitly to `::1`, it's going to end up listening on IPv4 only _or_ in dual-stack mode... so conversely, if our remote address resolves to a "mixed mode" set of addresses, that's a strong signal that we should be trying to use the IPv4 addresses in that list preferentially.

For cases when the user ends up binding to `::1` in the Agent (so IPv6 only) but configures their exporter to connect to `localhost` or whatever... that's kind of on them if `localhost` resolves to `[127.0.0.1, ::1]`... and given the statelessness of "connecting" to UDP, there's not a _whole_ lot we can do. 🤷🏻

PR #706, opened 2026-08-24
feat(metrics-tracing-context): add configurable per-field merge policy

arnaudbriche:
## What & why

`metrics-tracing-context` injects active span fields as metric labels. Today, when a child span defines a field that its parent also defines, the child's value simply **overrides** the parent's. That works for flat labels, but breaks for hierarchical values — e.g. a `component` path — where the useful label is the nested composition (`analyzer.worker.task`), not the innermost segment alone.

This PR makes the collision strategy configurable per field name, so hierarchical values can be composed from the span ancestry instead of collapsed.

## API

```rust
pub enum FieldMergePolicy {
    /// Child span overrides parent; recorded value overrides held. This is the
    /// historical behavior and the default.
    Override,
    /// Compose as `<outer><sep><inner>`: parent's value is outer, child's is inner.
    Append(&'static str),
}

MetricsLayer::new().with_field_merge_policy("component", FieldMergePolicy::Append("."));
```

With the above, given `component = "analyzer"` on the root span and `component = "worker"` on a nested span, metrics emitted within the nested span carry `component = "analyzer.worker"`. Three-level nesting yields `analyzer.worker.task`.

The policy applies at both places a field can collide:
- `on_new_span` — parent fields inherited into a child span
- `on_record` — `Span::record` setting a value for a field the span already holds (composes with the held value)

## Compatibility & performance

- **Backward compatible by default.** Any field name without an explicit policy uses `Override`, which is exactly the crate's current behavior. Existing behavior changes only where a policy is configured.
- **Zero cost on the metric hot path.** Policies are consulted only when a span is created/recorded, never during metric emission. When no policies are configured, the previous fast path is taken unchanged.

## Tests

Added to `tests/integration.rs`:
- Two- and three-level `Append` composition
- Gap in the chain (unannotated span) passes the composed value through
- `Span::record` composition, including cumulative re-records
- Metric-callsite labels still take precedence over composed span labels
- Per-field isolation (unconfigured fields keep `Override`)

## Out of scope

- No global/default policy (per-field only)
- Separator is `&'static str` (no runtime-computed separators)

## Note on CI

The Clippy job currently fails, but it is **unrelated to this change**: it is a pre-existing issue in `metrics-exporter-prometheus` (two `map_or(...)` identity expressions at `src/distribution.rs:155-156`) newly escalated to hard errors by the Rust 1.98.0 clippy `map_or_identity` lint, under that crate's `#![deny(clippy::all)]`. This PR only adds to `metrics-tracing-context`, which is clippy-clean.

### DIFF
--- a/metrics-tracing-context/src/field_merge.rs
+++ b/metrics-tracing-context/src/field_merge.rs
@@ -1,0 +1,32 @@
+//! Field merge policies.
+//!
+//! A [`FieldMergePolicy`] determines what happens when the same span field name shows up
+//! more than once: either a child span defines a field that its parent also defines, or
+//! [`Span::record`][tracing::Span::record] is used to set a value for a field that the span
+//! already holds. Policies are configured per field name via
+//! [`MetricsLayer::with_field_merge_policy`][crate::MetricsLayer::with_field_merge_policy].
+
+/// The merge policy applied when a value is merged over an existing value for the same
+/// span field name.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum FieldMergePolicy {
+    /// The most recently defined value wins: a child span's field overrides the value
+    /// inherited from its parent, and a recorded value overrides the previously held one.
+    ///
+    /// This is the historical behavior of the crate, and it applies to any field name
+    /// without an explicitly configured policy.
+    Override,
+    /// Compose both values hierarchically as `<outer><sep><inner>`.
+    ///
+    /// When a child span inherits a field from its parent, the parent's value is treated
+    /// as the outer value and the child's own value as the inner one. When a value is set
+    /// via [`Span::record`][tracing::Span::record], the previously held value is treated
+    /// as the outer value and the newly recorded value as the inner one. If only one of
+    /// the two values exists, it is used as-is.
+    ///
+    /// As spans nest, values accumulate into a path: given `component = "analyzer"` on the
+    /// root span, `component = "worker"` on a child span, and `component = "task"` on a
+    /// grandchild span, metrics emitted within the grandchild span see
+    /// `component = "analyzer.worker.task"`.
+    Append(&'static str),
+}
--- a/metrics-tracing-context/src/tracing_integration.rs
+++ b/metrics-tracing-context/src/tracing_integration.rs
@@ -1,10 +1,13 @@
 //! The code that integrates with the `tracing` crate.
 
+use crate::field_merge::FieldMergePolicy;
+use indexmap::map::Entry;
 use indexmap::IndexMap;
 use lockfree_object_pool::{LinearObjectPool, LinearOwnedReusable};
 use metrics::{Key, SharedString};
 use once_cell::sync::OnceCell;
 use std::cmp;
+use std::collections::HashMap;
 use std::sync::Arc;
 use tracing_core::span::{Attributes, Id, Record};
 use tracing_core::{field::Visit, Dispatch, Field, Subscriber};
@@ -95,17 +98,35 @@ impl AsRef<Map> for Labels {
 
 /// [`MetricsLayer`] is a [`tracing_subscriber::Layer`] that captures the span
 /// fields and allows them to be later on used as metrics labels.
+///
+/// Merge behavior for individual field names can be customized via
+/// [`MetricsLayer::with_field_merge_policy`].
 #[derive(Default)]
 pub struct MetricsLayer {
     #[allow(clippy::type_complexity)]
     with_labels:
         Option<fn(&Dispatch, &Id, f: &mut dyn FnMut(&Labels) -> Option<Key>) -> Option<Key>>,
+    field_merges: HashMap<String, FieldMergePolicy>,
 }
 
 impl MetricsLayer {
     /// Create a new [`MetricsLayer`].
     pub fn new() -> Self {
         Self::default()
+    }
+
+    /// Configures the [`FieldMergePolicy`] applied when merging values for the given
+    /// span field name.
+    ///
+    /// Field names without an explicitly configured policy are merged using
+    /// [`FieldMergePolicy::Override`], which preserves this crate's historical behavior.
+    pub fn with_field_merge_policy(
+        mut self,
+        field: impl AsRef<str>,
+        policy: FieldMergePolicy,
+    ) -> Self {
+        self.field_merges.insert(field.as_ref().to_owned(), policy);
+        self
     }
 
     pub(crate) fn with_labels(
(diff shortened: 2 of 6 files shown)


arnaudbriche:
Hi,

It seems like Clippy warnings are pre-existing.
@tobz first of all, any interest in this new feature ?

tobz:
@arnaudbriche 👋🏻 

I'll get those Clippy errors fixed and passing on `main` shortly.

Overall, yes: this seems like a very reasonable feature to add and I can understand the goal/use case. 👍🏻

metrics-tracing-context/src/field_merge.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// as the outer value and the child's own value as the inner one. When a value is set
+    /// via [`Span::record`][tracing::Span::record], the previously held value is treated
+    /// as the outer value and the newly recorded value as the inner one. If only one of
+    /// the two values exists, it is used as-is.
+    ///
+    /// As spans nest, values accumulate into a path: given `component = "analyzer"` on the
+    /// root span, `component = "worker"` on a child span, and `component = "task"` on a
+    /// grandchild span, metrics emitted within the grandchild span see
+    /// `component = "analyzer.worker.task"`.
+    Append(&'static str),
```
tobz: We should probably add a note here that this composition doesn't take effect for late-recorded parent fields, since they don't exist at the time where the append operation occurs.

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -152,7 +197,29 @@ where
...
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(v.clone());
+                        }
+                    },
+                    // `FieldMergePolicy::Override`: the newly recorded value wins.
+                    _ => {
+                        map.insert(k.clone(), v.clone());
+                    }
+                });
```
tobz: There's a small bug here around the fact that `on_record` could be called multiple times, and the main merge loop here extends the span's _existing_ set of labels (which might have already been append-merged in `on_new_span` or `on_record`), rather than recreating them recreating the current span's labels from the incoming record.

It would probably also be good to put that logic into a standalone function to share with the impl of `on_new_span`, as well.

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -139,7 +160,31 @@ where
...
+                } else {
+                    // Merging the parent's values over the child's own values: for
+                    // `Append`, the parent's value is the outer value and the value
+                    // already held by the child is the inner one.
+                    let field_merges = &self.field_merges;
+                    labels.extend(parent_labels, |map, k, v| {
+                        match field_merges.get(k.as_ref()) {
+                            Some(FieldMergePolicy::Append(sep)) => match map.entry(k.clone()) {
+                                Entry::Occupied(mut entry) => {
+                                    let composed = format!("{v}{sep}{}", entry.get());
```
tobz: In append mode, we should skip appending the value if it's completely empty, since you end up with something like `parent.` instead of the intended `parent.child`.

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -139,7 +160,31 @@ where
...
+                                    entry.insert(composed.into());
+                                }
+                                Entry::Vacant(entry) => {
+                                    entry.insert(v.clone());
+                                }
+                            },
+                            // `FieldMergePolicy::Override`: the child's own value wins.
+                            _ => {
```
tobz: Can we make this explicit so adding any new merge policy in the future doesn't silently drop down into this branch?

metrics-tracing-context/src/tracing_integration.rs
```diff
@@ -152,7 +197,29 @@ where
...
+                        Entry::Occupied(mut entry) => {
+                            let composed = format!("{}{sep}{v}", entry.get());
+                            entry.insert(composed.into());
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(v.clone());
+                        }
+                    },
+                    // `FieldMergePolicy::Override`: the newly recorded value wins.
+                    _ => {
```
tobz: Can we make this explicit so adding any new merge policy in the future doesn't silently drop down into this branch?

tobz (changes requested):
Apologies on the very delayed review, but left some feedback here.

arnaudbriche:
Thanks for the review, @tobz — all five comments are addressed in `75446cb` (the branch was also rebased onto the current `main`, `d9b5bf9`).

**`field_merge.rs` — note re: late-recorded parent fields.** The `Append` docs now call out that composition sees the parent chain as it exists at the moment of composition: a field that a parent records *after* a child was created can't appear in the child's already-composed value. (A later `Span::record` on the child *can* pick it up, since `on_record` re-reads the parent chain.)

**Empty-value guard (both call sites).** Composition is a shared helper that only composes when both sides are non-empty, so an empty value yields `parent`, not `parent.` (and never `.child`).

**`_ =>` catch-alls removed.** `merge_field` matches `Some(FieldMergePolicy::Append(sep))` and `Some(FieldMergePolicy::Override) | None` explicitly — the no-policy case is its own arm rather than a catch-all, so a future policy variant will fail to compile instead of silently falling into the `Override` branch.

**`on_record` — the re-record bug.** Agreed, appending onto the previously-composed labels was wrong. `on_record` now recomputes the span's labels for the recorded fields from the incoming record: the recorded value is composed with the value inherited from the parent chain (outer), and *replaces* the span's own previous contribution instead of accumulating onto it. Concretely, with `component` on `Append(".")`:
- parent `component = "analyzer"`, child records `worker` then `leaf` → `analyzer.worker` → `analyzer.leaf` (previously `analyzer.worker.leaf`), and
- re-recording the same value no longer double-counts: `analyzer.worker`, never `analyzer.worker.worker`.

This follows the "recreate from the incoming record" idea you described. One consequence to sign off on: a re-record *replaces* the span's own previous segment (from span creation or an earlier record), so recording a field is no longer cumulative. I think that's the right semantic, but I can flip it to accumulate if you'd rather.

The merge logic is now shared — both `on_new_span` and `on_record` delegate to a single `merge_field` (+ `compose_field_values`).

**Tests.** `test_field_merge_policy_record_composes` was updated for the replace semantics, and new tests were added: re-record doesn't double-count, empty values never produce dangling separators, and late-recorded parent fields don't leak into an already-created child's composition.

**CI note.** The pre-existing Clippy failures this PR originally hit (the `map_or`/redundant-cast lints in `metrics-exporter-prometheus` and `metrics`) have since landed on `main` (`9b827ae`, `d9b5bf9`). I rebased onto the refreshed main, and confirmed `cargo test -p metrics-tracing-context` (23 integration + 3 doc tests), `cargo fmt --check`, and `cargo clippy --workspace` are all clean locally.

PR #710, opened 2026-09-16
perf(metrics-util): stripe AtomicBucket to improve performance with concurrent writers

alex:
This is a series of 3 commits to improve performance when multiple threads write to the same metrics. Each commit has a detailed commit message describing what it's doing. The primary downside is that it increases memory usage -- sometimes significantly depending on the number of metrics and the number of threads who write to each metric. The top line performance numbers are impressive though:
```
22-core Xeon E5-2696 v4 (Linux):

  threads     before      after    speedup
        1    6.79 ms    6.00 ms      1.13x
        2   25.48 ms    6.49 ms      3.9x
        4   38.70 ms    6.92 ms      5.6x
        8   77.48 ms    8.02 ms      9.7x
       16  149.22 ms    7.28 ms     20.5x

10-core Apple M1 Max (macOS):

  threads     before      after    speedup
        1    1.31 ms    1.16 ms      1.13x
        2   14.58 ms    1.33 ms     11.0x
        4   42.66 ms    1.62 ms     26.3x
        8   99.62 ms    1.73 ms     57.6x
       16  251.66 ms    3.43 ms     73.4x
```


alex:
clippy failures look unrelated

alex:
If the memory increase is too big, I've got a potential follow up that changes the `Block` size to start smaller and then be larger for subsequent blocks -- no observed performance trade-off, and it reduces the memory overhead significantly. Happy to push that up as well, but I don't want to overload y'all with a zillion commits before anyone has had a change to take a look.

tobz:
@alex 👋🏻 Thanks for submitting this. Two things:

- I think it's fine to include the changes to use an incrementally larger block size since `AtomicBucket` can already be a little unwieldy to hold properly with regards to memory consumption. 😅
- Would it be possible to add in `loom` here? The concurrency model of what is happening here seems reasonable to me, but I hesitate to drop down closer to raw atomics from the otherwise battle-tested `crossbeam-epoch` without some compensating control like running it under `loom`.

alex:
Happy to! Will work on those this evening.

alex:
Ok, we've now got the size reduction, miri, and loom. This PR now has quite a bit of code, so if there's anything you want me to split out into it's own PR let me know.

tobz (commented):
Overall, this looks pretty reasonable to me.

I still need to take a clanker-assisted pass over some of the more thorny parts like the variable-sized block allocations, but conceptually I'm down with the design so it might just come down to trying to tweak the _approach_ to make it a little safer / easier to reason about, etc.

alex:
Sounds great, thanks for taking a look, and as I said, if it's helpful to split any of this out into seperate PRs just to make teh scope of the review easier, I'm happy to.

